Issue #5, opened 2025-04-18
GH Actions CI


mykola-mokhnach:
Module checks, like linter, build, tests, etc, must be automated and integrated with GitHub actions.
Also, all dependencies must be updated automatically via GH dependabot.

PR #12, opened 2025-04-25
CI: Added CI Job

saikrishna321:
Addressed #5


harsha509:
Hi @saikrishna321,

Could you please confirm if this PR is still valid? If it is no longer required, we may proceed with closing it.

Thanks,
Sri

PR #45, opened 2025-07-02
ci: add write permission for GITHUB_TOKEN

KazuCocoa:
Because this repository is newly created one recently

### DIFF
--- a/.github/workflows/publish.js.yml
+++ b/.github/workflows/publish.js.yml
@@ -8,6 +8,9 @@ on:
   push:
     branches: [ main ]
 
+permissions:
+  contents: write
+
 jobs:
   build:
     runs-on: ubuntu-latest


KazuCocoa:
Hm, this fixed the permission issue but it looks like the default bump by the semantics-release module bumped the version as 1.0.0. I don't know about the details, but maybe it needed to configure something 

https://github.com/appium/appium-ios-remotexpc/commit/ec47d4ad448b8da45f8d28cfb48aba2e51c972b6

@saikrishna321 

I think at least you could address this was under development in the README

KazuCocoa:
```
npm notice Publishing to https://registry.npmjs.org/ with tag latest and default access
npm error code E403
npm error 403 403 Forbidden - PUT https://registry.npmjs.org/appium-ios-remotexpc - appium-ios-remotexpc cannot be republished until 24 hours have passed.
npm error 403 In most cases, you or one of your dependencies are requesting
npm error 403 a package version that is forbidden by your security policy, or
npm error 403 on a server you do not have access to.
npm error A complete log of this run can be found in: /Users/kazu/.npm/_logs/2025-07-02T05_18_18_247Z-debug-0.log
```

So, I have published the 1.0.0. Then, to publish with 0.0.1, we should manually run the npm publish but it should be after 24hours. 

I'll disable the automatic release for now to avoid 1.0.0 release. Then, I'll manually publish the 0.0.1 after 24 hours manually @saikrishna321

PR #51, opened 2025-07-04
feat: add SRP implementation for Apple TV authentication

harsha509:
### Description

This module implements the **Secure Remote Password (SRP-6a)** protocol for Apple TV device pairing.

It provides zero-knowledge proof authentication that allows secure pairing using a PIN code without transmitting the PIN over the network.

### SRP is needed for:

• Apple TV requires SRP for remote control pairing
• Authenticates using PIN without transmitting it
• No pre-shared secrets needed
• Provides mutual authentication
• Generates session keys for encryption
• PIN-based user experience

### DIFF
--- a/src/lib/apple-tv/srp/srp-client.ts
+++ b/src/lib/apple-tv/srp/srp-client.ts
@@ -89,48 +89,58 @@ export class SRPClient {
   }
 
   /**
+   * Gets the salt value received from the server.
+   *
+   * @returns The salt buffer or null if not set
+   */
+  get salt(): Buffer | null {
+    return this._salt;
+  }
+
+  /**
    * Sets the salt value received from the server.
    *
-   * @param salt - The salt buffer from the server
+   * @param value - The salt buffer from the server
    * @throws {SRPError} If salt is empty or client is disposed
    */
-  public setSalt(salt: Buffer): void {
+  set salt(value: Buffer) {
     this.throwIfDisposed();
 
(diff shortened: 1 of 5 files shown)


src/lib/apple-tv/srp/srp-client.ts
```diff
@@ -0,0 +1,369 @@
...
+    log.debug('Identity set successfully');
+  }
+
+  /**
+   * Sets the salt value received from the server.
+   *
+   * @param salt - The salt buffer from the server
+   * @throws {SRPError} If salt is empty or client is disposed
+   */
+  public setSalt(salt: Buffer): void {
```
mykola-mokhnach, line 97: should we instead use setters and getters? (e.g. `set salt(value)/get salt()` )

mykola-mokhnach (approved):

src/lib/apple-tv/srp/srp-client.ts
```diff
@@ -0,0 +1,393 @@
...
```
mykola-mokhnach, line 97: I assume the payload of this function should be moved to the setter. There is no need to have two public setters doing the same stuff

mykola-mokhnach: same below

harsha509: Thank you for pointing this out. 
I have now corrected the oversight.

PR #63, opened 2025-08-26
fix: use 0.9 explicitly

KazuCocoa:
It looks like the library released [0.8.11](https://www.npmjs.com/package/@xmldom/xmldom/v/0.8.11) recently, but the code in this repository expects to use 0.9 (e.g. https://www.npmjs.com/package/@xmldom/xmldom/v/0.9.8)

The v0.9 includes Element, for example, properly so we should explicitly require `0.9+`

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,7 +69,7 @@
     "@appium/strongbox": "^1.0.0-rc.1",
     "@appium/support": "^7.0.0-rc.1",
     "@types/node": "^24.0.10",
-    "@xmldom/xmldom": "^0.x",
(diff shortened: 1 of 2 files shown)


KazuCocoa:
- `/@xmldom/xmldom/index.d.ts` https://www.npmjs.com/package/@xmldom/xmldom/v/0.9.8?activeTab=code

vs

- `/@xmldom/xmldom/index.d.ts` https://www.npmjs.com/package/@xmldom/xmldom/v/0.8.11?activeTab=code

Issue #70, opened 2025-09-13
Implement MCInstall service


Delta456:
MCInstall service handles installation, listing and deleting Apple profiles.

This service is a must implement for `remotexpc`

Pymobiledevice implementation can be found at https://github.com/doronz88/pymobiledevice3/blob/master/pymobiledevice3/services/mobile_config.py

PR #71, opened 2025-09-15
feat: implement `mcinstall` service

Delta456:
Implements shim `MCInstall` service for new iOS versions

closes https://github.com/appium/appium-ios-remotexpc/issues/70

### DIFF
--- a/src/services/ios/mobile-config/index.ts
+++ b/src/services/ios/mobile-config/index.ts
@@ -77,7 +77,7 @@ class MobileConfigService
   /**
    * Install profile from path to iOS device
    * The phone must be unlocked before installing the profile
-   * @param {String} path  must be a certificate file .PEM .CER and more formats
+   * @param {String} filePath  must be a certificate file .PEM .CER and more formats
    * e.g: /Downloads/charles-certificate.pem
    */
   async installProfileFromPath(filePath: string): Promise<void> {
@@ -88,8 +88,11 @@ class MobileConfigService
       throw new Error(`Profile filepath does not exist: ${filePath}`);
     }
 
-    const fileExtension = path.extname(filePath);
-    if (!fileExtension || !SUPPORTED_EXTENSIONS.includes(`.${fileExtension}`)) {
+    const fileExtension = path.extname(filePath).toLowerCase();
+    if (
+      !fileExtension ||
+      !SUPPORTED_EXTENSIONS.includes(fileExtension as any)
+    ) {
       throw new Error(
         `Unsupported file format. Supported formats: ${SUPPORTED_EXTENSIONS.join(', ')}`,
       );
@@ -174,22 +177,12 @@ class MobileConfigService
     await this._conn.sendAndReceive(req);
     const res = await this._conn.sendAndReceive(req);
     if (res.Status !== 'Acknowledged') {
-      const errorChain = res.ErrorChain;
-      if (Array.isArray(errorChain) && errorChain.length > 0) {
-        const firstError = errorChain[0];
-        if (
-          typeof firstError === 'object' &&
-          firstError !== null &&
-          'ErrorCode' in firstError
-        ) {
-          const errorCode = (firstError as any).ErrorCode;
-          if (Number.isInteger(errorCode)) {
-            if (errorCode === ERROR_CLOUD_CONFIGURATION_ALREADY_PRESENT) {
-              throw new Error(
-                'A cloud configuration is already present on device. You must first erase the device to install a new configuration.',
-              );
-            }
-          }
+      const errorCode = (res.ErrorChain as any[])?.[0]?.ErrorCode;
+      if (Number.isInteger(errorCode)) {
+        if (errorCode === ERROR_CLOUD_CONFIGURATION_ALREADY_PRESENT) {
+          throw new Error(
+            'A cloud configuration is already present on device. You must first erase the device to install a new configuration.',
+          );
         }
       }
       throw new Error(`Invalid response: ${JSON.stringify(res)}`);
--- a/src/lib/plist/constants.ts
+++ b/src/lib/plist/constants.ts
@@ -61,4 +61,4 @@ export const SUPPORTED_EXTENSIONS = [
   '.pfx',
   '.mobileconfig',
   '.plist',
-];
+] as const;


SrinivasanTarget (changes requested):

src/lib/types.ts
```diff
@@ -212,6 +212,52 @@ export interface NotificationProxyService extends BaseService {
...
+   *   Status: 'Acknowledged'
+   * }
+   */
+  getProfileList(): Promise<PlistDictionary>;
+  /**
+   * Install profile to iOS device
+   * @param {String} path  must be a certificate file .PEM .CER and more formats
+   * e.g: /Downloads/charles-certificate.pem
+   */
+  installProfile(path: string): Promise<void>;
```
mykola-mokhnach, line 253: there should also be an option to install a profile from a memory buffer

src/services/ios/mobile-config/index.ts
```diff
@@ -0,0 +1,171 @@
...
+      return this._conn;
+    }
+
+    const service = this.getServiceConfig();
+    this._conn = await this.startLockdownService(service);
+    return this._conn;
+  }
+
+  /**
+   * Get all profiles of iOS devices
```
mykola-mokhnach, line 46: we should not copy docstrings if they are already present in the interface declaration

src/services/ios/mobile-config/index.ts
```diff
@@ -0,0 +1,171 @@
...
+    if (!this._conn) {
+      this._conn = await this.connectToMobileConfigService();
+    }
+
+    const req = {
+      RequestType: 'RemoveProfile',
+      ProfileIdentifier: createPlist(payloadData),
+    };
+
+    log.info(req);
```
mykola-mokhnach, line 120: do we need this log line?

src/services/ios/mobile-config/index.ts
```diff
@@ -0,0 +1,171 @@
...
+    await this._sendPlistAndReceive(req);
+  }
+
+  private async _sendPlistAndReceive(
+    req: PlistDictionary,
+  ): Promise<PlistDictionary> {
+    if (!this._conn) {
+      this._conn = await this.connectToMobileConfigService();
+    }
+    // Ignore first response as it is just status check
```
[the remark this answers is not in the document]
Delta456: Not sure. We need to ignore the first response.

Delta456:
> > @mykola-mokhnach If possible I need help with `removeProfile`. The plist request is sent and gets acknowledged but the profile is not removed. I don't know whats the issue 😓
> 
> You could also try checking if the plist encoding is correct. I previously found that the plist implementation wasn't correct.

The plist encoding is correct. I have cross checked with `pymobiledevice3`.

src/services/ios/mobile-config/index.ts
Delta456: It was a log for debugging why `deleteProfile` is not working. I still need to figure it out.

src/services/ios/mobile-config/index.ts
mykola-mokhnach: I think this is redundant and must be fixed. Only the interface is visible to the "outside" and should contain docs. Implementations are private to the module though.
WDYT @saikrishna321 @SrinivasanTarget  ?

src/services/ios/mobile-config/index.ts
```diff
@@ -0,0 +1,217 @@
...
+      throw error;
+    }
+  }
+  /**
+   * Remove profile from iOS device
+   * @param {String} identifier Query identifier list through getProfileList method
+   */
+  async removeProfile(identifier: string): Promise<void> {
+    const profileList = await this.getProfileList();
+    if (!profileList || !profileList.ProfileMetadata) {
```
mykola-mokhnach, line 137: could be simplified  to `if (!profileList?.ProfileMetadata)`

src/services/ios/mobile-config/index.ts
```diff
@@ -0,0 +1,217 @@
...
+      return;
+    }
+
+    const profileMetadata = profileList.ProfileMetadata as Record<string, any>;
+    if (!(identifier in profileMetadata)) {
+      // Get available identifiers from OrderedIdentifiers array or ProfileMetadata keys
+      const availableIdentifiers =
+        profileList.OrderedIdentifiers &&
```
mykola-mokhnach, line 145: Consider using explicit if/else instead. This expression might result in something that you don't expect if `profileList.OrderedIdentifiers` is falsy/undefined

src/services/ios/mobile-config/index.ts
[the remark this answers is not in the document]
mykola-mokhnach: does it work the same with pymobiledevice? that's weird

src/services/ios/mobile-config/index.ts
```diff
@@ -0,0 +1,217 @@
...
+    req: PlistDictionary,
+  ): Promise<PlistDictionary> {
+    if (!this._conn) {
+      this._conn = await this.connectToMobileConfigService();
+    }
+    // Ignore first response as it is just status check
+    await this._conn.sendAndReceive(req);
+    const res = await this._conn.sendAndReceive(req);
+    if (res.Status !== 'Acknowledged') {
+      const errorChain = res?.ErrorChain;
```
mykola-mokhnach, line 183: the previous condition already ensures that `res` is not falsy, so optional chaining here is redundant

src/services/ios/mobile-config/index.ts
```diff
@@ -0,0 +1,171 @@
...
+    if (res.Status !== 'Acknowledged') {
+      const errorChain = res.ErrorChain;
+      if (
+        errorChain != null &&
+        Array.isArray(errorChain) &&
+        errorChain.length > 0
+      ) {
+        const firstError = errorChain[0];
+        if (
+          typeof firstError === 'object' &&
```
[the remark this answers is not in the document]
mykola-mokhnach: not sure if it's fixed, the condition is still present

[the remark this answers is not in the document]
Delta456: I get `TS7053: Element implicitly has an any type because expression of type 0 can't be used to index type`

src/lib/types.ts
```diff
@@ -212,6 +212,57 @@ export interface NotificationProxyService extends BaseService {
...
+   *   Status: 'Acknowledged'
+   * }
+   */
+  getProfileList(): Promise<PlistDictionary>;
+  /**
+   * Install profile to iOS device
+   * @param {String} path  must be a certificate file .PEM .CER and more formats
+   * e.g: /Downloads/charles-certificate.pem
+   */
+  installProfileFromPath(path: string): Promise<void>;
```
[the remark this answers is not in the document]
mykola-mokhnach: fine with me

src/services/ios/mobile-config/index.ts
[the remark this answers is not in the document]
Delta456: Done

src/services/ios/mobile-config/index.ts
Delta456: We can do this later.

mykola-mokhnach (approved):

PR #74, opened 2025-10-03
feat(index): export Notification and MobileConfig

Delta456:
Export NotificationProxy and MobileConfig with their connections for driver usage


harsha509 (approved):
LGTM!

PR #75, opened 2025-10-04
fix(plist): enforce strict length‑prefixed framing in `LengthBasedSplitter`

navin772:
Some RSD services (WebInspector, springboard) intermittently returned no messages and timed out on receive.

After debugging I found out that the `LengthBasedSplitter` emitted raw XML buffers without the 4‑byte length header which caused no response to be received/parsed from the service.

The plist decoder would attempt to slice off a header that wasn’t there, producing malformed XML. Due to this the response was empty and `receivePlist` timed out.

This issue was found out while implementing the webinspector and springboard service, both of which timed out, and this fix resolved the issue.

Also, we need to use `unified-plist-parser` in plist decoder to auto detect between binary and XML.

I have tested other services with this change and there is no regression.

### DIFF
--- a/src/lib/plist/length-based-splitter.ts
+++ b/src/lib/plist/length-based-splitter.ts
@@ -100,14 +100,6 @@ export class LengthBasedSplitter extends Transform {
         Math.min(MAX_PREVIEW_LENGTH, this.buffer.length),
       );
 
-      // Check for XML format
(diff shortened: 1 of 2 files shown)


harsha509 (approved):
There was a reason I did that, something to do with testing out syslogs, but I can’t quite recall why!

Looks good to me though, XML should have the 4-byte header too.

Good catch, @navin772 !

Issue #77, opened 2025-10-06
implement springboard service


Delta456:
`springboard` is useful for getting getting iconData, IconState, wallpapers and several other things

`pymobiledevice3` can act as a source of inspiration as it already implements this [service](https://github.com/doronz88/pymobiledevice3/blob/master/pymobiledevice3/cli/springboard.py)

PR #79, opened 2025-10-07
implement springboard service 

Delta456:
This PR implements springboard service which works through RSD `com.apple.springboardservices.shim.remote`. 

This service is used for getting iconData, IconState, wallpapers and several other things

Fixes issue https://github.com/appium/appium-ios-remotexpc/issues/77

Note: `setIconState`, `getWallpaperInfo` and `getWallpaperPNGData` don't work due to a bug in Apple's protocol. This is already being tracked through [`pymobildevice3`](https://github.com/doronz88/pymobiledevice3/issues/1450)

### DIFF
--- a/src/lib/types.ts
+++ b/src/lib/types.ts
@@ -6,6 +6,7 @@ import { EventEmitter } from 'events';
 
 import type { ServiceConnection } from '../service-connection.js';
 import type { BaseService, Service } from '../services/ios/base-service.js';
+import type { InterfaceOrientation } from '../services/ios/springboard-service/index.js';
 import type { RemoteXpcConnection } from './remote-xpc/remote-xpc-connection.js';
 import type { Device } from './usbmux/index.js';
 
@@ -519,6 +520,8 @@ export interface SpringboardService extends BaseService {
    */
   getIconState(): Promise<PlistDictionary>;
 
+  setIconState(newState: PlistDictionary[]): Promise<void>;
+
   /**
    * Gets the icon PNG data for a given bundle ID
    * @param bundleID The bundle ID of the app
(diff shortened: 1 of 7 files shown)


src/lib/types.ts
```diff
@@ -508,3 +509,78 @@ export interface MobileImageMounterServiceWithConnection {
...
+   * @returns Promise resolving to the icon state
+   */
+  getIconState(): Promise<PlistDictionary>;
+
+  /**
+   * TODO: This does not work currently due to a bug in Apple protocol implementation (maybe?)
+   * Sets the icon state
+   * @param newState
+   */
+  setIconState(newState: PlistDictionary[]): Promise<void>;
```
mykola-mokhnach, line 528: Not sure if we need these methods if they don't really work...

mykola-mokhnach: Also, it does not really make sense to have such method without detailed argument payload documentation. If I don't know what should be passed to it and it's not documented anywhere then I obviously cannot use it.

src/services/ios/springboard-service/index.ts
```diff
@@ -0,0 +1,202 @@
...
+  type PlistDictionary,
+  type SpringboardService as SpringboardInterface,
+} from '../../../lib/types.js';
+import { ServiceConnection } from '../../../service-connection.js';
+import { BaseService } from '../base-service.js';
+
+enum InterfaceOrientation {
+  PORTRAIT = 1,
+  PORTRAIT_UPSIDE_DOWN = 2,
+  LANDSCAPE = 3,
```
mykola-mokhnach, line 11: perhaps it would also make sense to mention  which of these is 90 grad and which one is 270

src/services/ios/springboard-service/index.ts
```diff
@@ -0,0 +1,202 @@
...
+      return await this.sendRequestAndReceive(req);
+    } catch (error) {
+      if (error instanceof Error) {
+        throw new Error(`Failed to get wallpaper info: ${error.message}`);
+      }
+      throw error;
+    }
+  }
+
+  async getWallpaperPreviewImage(wallpaperName: string): Promise<Buffer> {
```
mykola-mokhnach, line 94: we may also enforce this limitation with types: `wallpaperName: 'homescreen' | 'lockscreen'`

PR #82, opened 2025-10-07
chore(tss): remove unnecessary type conversion

navin772:
Axios handles string and buffer automatically.

`createPlist(this._request)` returns string (XML) and we can send it directly to TSS server via axios.

Tested on device and is working fine.


harsha509 (approved):
LGTM

PR #87, opened 2025-10-21
feat: add power assertion agent shim service

### LINKED ISSUE #86: Implement power assertion agent RSD service
RSD service name - `com.apple.mobile.assertion_agent.shim.remote`

pymobiledevice3 implementation for reference - https://github.com/doronz88/pymobiledevice3/blob/master/pymobiledevice3/services/power_assertion.py

navin772:
Fixes #86 

I have added only unit tests since integration tests are out of scope for this service. There is no testable way that is currently supported.

Also, do note that power assertion doesn't overrides your device display sleep and is related to CPU sleep.

For verification on whether the service works or not, I have manually tested this out by capturing syslogs at the time of running the power assertion.

For example, before running the `createPowerAssertion` method, start the live syslogs - `sudo pymobiledevice3 syslog live | grep "assertion"` and grep for `assertion`, you will see related logs along with the `timeout` you passed.

### DIFF
--- a/src/services/ios/power-assertion/index.ts
+++ b/src/services/ios/power-assertion/index.ts
@@ -12,7 +12,7 @@ const log = logger.getLogger('PowerAssertionService');
 /**
  * Power assertion types that can be used to prevent system sleep
  */
-enum PowerAssertionType {
+export enum PowerAssertionType {
   WIRELESS_SYNC = 'AMDPowerAssertionTypeWirelessSync',
   PREVENT_USER_IDLE_SYSTEM_SLEEP = 'PreventUserIdleSystemSleep',
   PREVENT_SYSTEM_SLEEP = 'PreventSystemSleep',
@@ -97,4 +97,4 @@ class PowerAssertionService
   }
 }
 
-export { PowerAssertionService, PowerAssertionType };
+export { PowerAssertionService };


src/services/ios/power-assertion/index.ts
```diff
@@ -0,0 +1,110 @@
...
+} from '../../../lib/types.js';
+import { ServiceConnection } from '../../../service-connection.js';
+import { BaseService } from '../base-service.js';
+
+const log = logger.getLogger('PowerAssertionService');
+
+/**
+ * Power assertion types that can be used to prevent system sleep
+ */
+enum PowerAssertionType {
```
mykola-mokhnach: should this enum be properly exported?

src/services/ios/power-assertion/index.ts
```diff
@@ -0,0 +1,110 @@
...
+class PowerAssertionService
+  extends BaseService
+  implements PowerAssertionServiceInterface
+{
+  static readonly RSD_SERVICE_NAME =
+    'com.apple.mobile.assertion_agent.shim.remote';
+
+  private _conn: ServiceConnection | null = null;
+
+  constructor(address: [string, number]) {
```
mykola-mokhnach: constructor is not needed if everything it does is to call super

src/lib/types.ts
```diff
@@ -213,6 +213,32 @@ export interface NotificationProxyService extends BaseService {
...
+  /**
+   * Create a power assertion to prevent system sleep
+   * @param type The type of power assertion to create
+   * @param name A descriptive name for the assertion
+   * @param timeout Timeout in seconds for how long the assertion should last
+   * @param [details] Additional details about the assertion
+   * @returns Promise that resolves when the assertion is created
+   */
+  createPowerAssertion(
+    type: string,
```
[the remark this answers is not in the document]
navin772: Do we get any additional benefits after doing that? I think it is fine this way since the params are fixed with only 1 optional arg

src/services/ios/power-assertion/index.ts
mykola-mokhnach: the thing is that the actual method implementation should not differ from its interface declaration. enums are only used to improve typechecks. they make no sense in the "raw" javascript code because it "erases" them to strings anyway

src/lib/types.ts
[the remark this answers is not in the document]
mykola-mokhnach: https://www.codu.co/articles/why-objects-are-my-go-to-for-handling-multiple-parameters-in-javascript-functions-wws-68kv
https://www.servicenow.com/community/developer-blog/coding-best-practices-the-power-of-passing-objects-as-parameters/ba-p/3234855
https://www.baeldung.com/java-best-practices-many-parameters-method
https://www.geeksforgeeks.org/javascript/difference-between-multiple-arguments-and-options-object
https://www.ericfeminella.com/blog/2013/01/22/simplifying-designs-with-parameter-objects/

src/services/ios/power-assertion/index.ts
navin772: Ok, I will use `PowerAssertionType` ENUM for type check in the interface declaration

src/lib/types.ts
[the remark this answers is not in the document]
navin772: Done, I added a interface - `PowerAssertionOptions`

Now, usage looks like:

```ts
await powerAssertionService.createPowerAssertion({
      type: PowerAssertionType.WIRELESS_SYNC,
      name: assertionName,
      timeout: 3,
      details: assertionDetails,
    });
```

navin772:
I have added 2 integration tests so we are atleast testing out power assertion against the device and see if it accepts it without any errors. 
I have verified that power assertion is logged in the live syslogs.

navin772:
@mykola-mokhnach updated the PR as per above discussions, could you please re-review it?

Issue #83, opened 2025-10-10
implement RSD support for `misagent`


Delta456:
The `misagent` service allows installation, removal and listing of `.mobileprovision` file

`pymobiledevice3` has an implementation which can be used for inspiration https://github.com/doronz88/pymobiledevice3/blob/master/pymobiledevice3/services/misagent.py

PR #88, opened 2025-10-21
feat: implement misagent

Delta456:
Implements `misagent` service support

Closes https://github.com/appium/appium-ios-remotexpc/issues/83

### DIFF
--- a/src/lib/types.ts
+++ b/src/lib/types.ts
@@ -796,7 +796,7 @@ export interface MisagentService extends BaseService {
   removeProfile(uuid: string): Promise<void>;
 
   /**
-   * Copies all provisioning profiles from the device
+   * Fetching all provisioning profiles from the device
    * This can be used for listing installed provisioning profiles and backing them up
    * @returns {Promise<ProvisioningProfile[]>}
    * e.g.
@@ -826,14 +826,14 @@ export interface MisagentService extends BaseService {
    * },
    * ]
    * @example
-   * const profiles = await misagentService.copyAll();
+   * const profiles = await misagentService.fetchAll();
    * profiles.forEach(profile => {
    *   console.log(`Profile: ${profile.plist.Name}`);
    *   console.log(`UUID: ${profile.plist.UUID}`);
    *   console.log(`Expires: ${profile.plist.ExpirationDate}`);
    * });
    */
-  copyAll(): Promise<ProvisioningProfile[]>;
+  fetchAll(): Promise<ProvisioningProfile[]>;
 }
 
 export interface MisagentServiceWithConnection {
--- a/src/services/ios/misagent/index.ts
+++ b/src/services/ios/misagent/index.ts
@@ -60,7 +60,7 @@ class MisagentService extends BaseService implements MisagentServiceInterface {
     );
   }
 
-  async copyAll(): Promise<ProvisioningProfile[]> {
+  async fetchAll(): Promise<ProvisioningProfile[]> {
     const response = await this.sendRequest(
       {
         MessageType: 'CopyAll',
--- a/test/integration/misagent-service-test.ts
+++ b/test/integration/misagent-service-test.ts
@@ -47,7 +47,7 @@ describe('MisagentService', function () {
   describe('copyAll', function () {
     it('should copy all installed profiles', async function () {
       try {
-        const res = await misagentService.copyAll();
+        const res = await misagentService.fetchAll();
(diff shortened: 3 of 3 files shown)


src/lib/types.ts
```diff
@@ -734,3 +735,62 @@ export interface SpringboardServiceWithConnection {
...
   remoteXPC: RemoteXpcConnection;
 }
+
+/**
+ * Represents the instance side of MisagentService
+ */
+export interface MisagentService extends BaseService {
+  /**
+   * Installs a provisioning profile from a file path
+   * @param path The file path of the provisioning profile
```
mykola-mokhnach: shall we add more details about what `provisioning profile` is? for example that it's required that the file has a specific extension/format?

src/services/ios/misagent/index.ts
```diff
@@ -0,0 +1,116 @@
...
+
+    const service = this.getServiceConfig();
+    this._conn = await this.startLockdownService(service);
+    return this._conn;
+  }
+
+  async installProfileFromPath(filePath: string): Promise<void> {
+    // Check if file exists
+    try {
+      await fs.access(filePath);
```
mykola-mokhnach: prefer using `fs.exists`

src/lib/types.ts
Delta456: Definitely! I will add details in a later commit

src/services/ios/misagent/index.ts
Delta456: We use `fs.access` here https://github.com/appium/appium-ios-remotexpc/blob/main/src/services/ios/mobile-config/index.ts#L86

src/services/ios/misagent/index.ts
```diff
@@ -0,0 +1,116 @@
...
+      return res;
+    } catch (error) {
+      if (error instanceof Error) {
+        throw new Error(`${errorMessage}: ${error.message}`);
+      }
+      throw error;
+    }
+  }
+
+  private getServiceConfig(): { serviceName: string; port: string } {
```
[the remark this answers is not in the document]
Delta456: I don't see the need because this declaration is only used once.

src/services/ios/misagent/provisioning-profile.ts
```diff
@@ -0,0 +1,47 @@
...
+    // Find '<?xml' marker and extract everything after it
+    const xmlMarker = Buffer.from('<?xml');
+    const xmlIndex = buf.indexOf(xmlMarker);
+
+    if (xmlIndex === -1) {
+      throw new Error('No XML content found in provisioning profile');
+    }
+
+    // Get XML content starting from '<?xml'
+    let xml = Buffer.concat([
```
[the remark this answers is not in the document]
Delta456: I will try to see if it can be simplified!

src/lib/types.ts
```diff
@@ -768,3 +769,74 @@ export interface SpringboardServiceWithConnection {
...
+   * ]
+   * @example
+   * const profiles = await misagentService.copyAll();
+   * profiles.forEach(profile => {
+   *   console.log(`Profile: ${profile.plist.Name}`);
+   *   console.log(`UUID: ${profile.plist.UUID}`);
+   *   console.log(`Expires: ${profile.plist.ExpirationDate}`);
+   * });
+   */
+  copyAll(): Promise<ProvisioningProfile[]>;
```
mykola-mokhnach: how about renaming this method to list or listAll?

src/services/ios/misagent/index.ts
```diff
@@ -0,0 +1,110 @@
...
+    return response.Payload.filter((profileData): profileData is Buffer =>
+      Buffer.isBuffer(profileData),
+    ).map((profileData) => new ProvisioningProfile(profileData));
+  }
+
+  private async sendRequest(
+    request: any,
+    errorMessage: string,
+  ): Promise<PlistDictionary> {
+    const conn = await this.connectToMisagentService();
```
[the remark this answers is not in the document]
Delta456: DIsconnect as in? I don't think we need to do that in general

src/lib/types.ts
Delta456: Prefect! `fetchAll` sounds closer to the current behavior

Delta456: should be done now!

src/services/ios/misagent/index.ts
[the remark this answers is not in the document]
Delta456: I checked again we cannot "disconnect" a service.

Issue #101, opened 2025-11-19
Add support for various DVT instruments


navin772:
DVT supports various instruments that we could use over the DVT RSD service via channels, implement these to extend support in appium-remotexpc:

- [x] Location Simulation - https://github.com/appium/appium-ios-remotexpc/pull/94
- [x] Condition Inducer - https://github.com/appium/appium-ios-remotexpc/pull/94
- [x] DeviceInfo - https://github.com/appium/appium-ios-remotexpc/pull/105
- [x] ApplicationListing - https://github.com/appium/appium-ios-remotexpc/pull/103
- [x] Screenshot - https://github.com/appium/appium-ios-remotexpc/pull/102
- [x] ProcessControl - https://github.com/appium/appium-ios-remotexpc/pull/136
- [x] Graphics - https://github.com/appium/appium-ios-remotexpc/pull/104
- [x] NetworkMonitor - https://github.com/appium/appium-ios-remotexpc/pull/110
- [ ] Sysmontap
- [ ] EnergyMonitor
- [x] Notifications - https://github.com/appium/appium-ios-remotexpc/pull/112
- [ ] ActivityTraceTap

Some instruments from above are streaming and TAP based and will require some modifications in DVT.

navin772:
@saikrishna321 here's the PR to add support for condition inducer to xcuitest driver via remotexpc - https://github.com/appium/appium-xcuitest-driver/pull/2657

Delta456:
This issue can be closed for now as all available DVT instruments are implemented.

PR #103, opened 2025-11-21
feat(dvt): implement app list

Delta456:
Implements ApplicationListing from DVT RSD service, part of https://github.com/appium/appium-ios-remotexpc/issues/101

Minor changes to nskeyed decoder to prevent depth limit and stack overflow

### DIFF
--- a/test/integration/dvt_instruments/app-listing-test.ts
+++ b/test/integration/dvt_instruments/app-listing-test.ts
@@ -38,7 +38,7 @@ describe('Application Listing', function () {
       expect(list).to.be.an('array');
       expect(list).to.not.be.empty;
 
-      expect(list.length).to.greaterThan(100);
+      expect(list.length).to.greaterThan(0);
       list.forEach((app) => {
         expect(app).to.have.property('CFBundleIdentifier');
         expect(app).to.have.property('DisplayName');


src/services/ios/dvt/nskeyedarchiver-decoder.ts
```diff
@@ -79,53 +79,115 @@ export class NSKeyedArchiverDecoder {
...
+          } else {
+            result[key] = value;
+          }
+        }
+
+        this.decoded.set(index, result);
+        visited.delete(index);
+        return result;
+      }
+    }
```
navin772: I don't think we require this, it just appends `{_isNSError: true}` to response in case of NSError, other decoding done here is already handled.

src/services/ios/dvt/nskeyedarchiver-decoder.ts
```diff
@@ -79,53 +79,115 @@ export class NSKeyedArchiverDecoder {
...
    * Decode an object at a specific index
    */
-  private decodeObject(index: number): any {
+  private decodeObject(
+    index: number,
+    visited: Set<number> = new Set(),
+    depth: number = 0,
+  ): any {
+    // Prevent stack overflow with depth limit
+    if (depth > 1000) {
```
mykola-mokhnach: please move max depth into a constant

mykola-mokhnach (approved):

PR #126, opened 2026-01-04
feat: add installation-proxy support service

harsha509:
Current behavior

* If the app is not installed, it performs a fresh install
* If the same version is already installed, the install is skipped
* If an older version is installed and the IPA is newer, the app is upgraded
* If a newer version is already installed, the operation is skipped (downgrades not supported)

### DIFF
--- a/src/services/ios/installation-proxy/index.ts
+++ b/src/services/ios/installation-proxy/index.ts
@@ -9,8 +9,6 @@ import type {
   ApplicationType,
   BrowseOptions,
   BrowseResponse,
-  InstallAction,
-  InstallOperationResult,
   InstallOptions,
   LookupOptions,
   LookupResponse,
@@ -18,18 +16,6 @@ import type {
   ProgressResponse,
   UninstallOptions,
 } from './types.js';
-
-/**
- * Context object to bundle install/upgrade operation parameters
- */
-interface InstallContext {
-  bundleIdentifier: string;
-  packagePath: string;
-  targetVersion: string;
-  installOptions: InstallOptions;
-  progressCallback?: ProgressCallback;
-  currentVersion?: string;
-}
 
 const log = getLogger('InstallationProxyService');
 
@@ -61,18 +47,6 @@ export const MAX_BROWSE_DURATION_MS = 2 * 60 * 1000; // 2 minutes
  * Safety limit to prevent endless loops while allowing time for large apps
  */
 export const MAX_INSTALL_DURATION_MS = 10 * 60 * 1000; // 10 minutes
-
-/**
- * Constants for install/upgrade operation messages
- */
-const INSTALL_MESSAGES = {
-  FRESH_INSTALL: 'App was not previously installed',
-  VERSION_UNKNOWN: 'Current version could not be determined',
-  DOWNGRADE_NOT_SUPPORTED:
-    'Downgrades are not supported by iOS installation_proxy',
-  SAME_VERSION_NOT_SUPPORTED:
-    'Reinstalling the same version is not supported by iOS installation_proxy',
-} as const;
 
 /**
  * InstallationProxyService provides an API to manage app installation and queries
@@ -250,7 +224,7 @@ export class InstallationProxyService extends BaseService {
     const request: PlistDictionary = {
       Command: 'Uninstall',
       ApplicationIdentifier: bundleIdentifier,
-      ClientOptions: options,
+      ClientOptions: options as PlistDictionary,
     };
 
     await this.executeWithProgress(request, progressCallback);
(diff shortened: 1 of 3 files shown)


src/services/ios/installation-proxy/types.ts
Delta456: Exporting each type linewise looks very redudant. 

We should go with `export { ApplicationType, BrowserOptions, ...}` which is the same convention we do for this project

src/services/ios/installation-proxy/constants.ts
Delta456: Such low amount of consts can easily be moved to `index.ts`

harsha509: i think, i should better move all constants to index.ts, as these are only constants and would not grow!!

src/services/ios/installation-proxy/index.ts
```diff
@@ -0,0 +1,342 @@
...
+
+      this.checkForError(response);
+
+      // Report progress if available
+      if (response.PercentComplete !== undefined && response.Status) {
+        const percent = response.PercentComplete;
+        const status = response.Status;
+        log.debug(`Progress: ${percent}% - ${status}`);
+
+        progressCallback?.(percent, status);
```
mykola-mokhnach: must this callback be synchronous?

src/services/ios/installation-proxy/index.ts
```diff
@@ -0,0 +1,342 @@
...
+      }
+
+      // Only break when we receive the final "Complete" status without progress info
```
mykola-mokhnach: may there be a situation the downstream API never returns this status and the loop is stuck?

src/services/ios/installation-proxy/index.ts
```diff
@@ -0,0 +1,342 @@
...
+      }
+    }
+
+    return result;
+  }
+
+  /**
+   * Lookup application information by bundle IDs
+   */
+  async lookup(
```
mykola-mokhnach: is it possible to add a helper to figure out if a particular app is already installed?

test/unit/installation-proxy/installation-proxy-service.spec.ts
```diff
@@ -0,0 +1,43 @@
+import { expect } from 'chai';
```
mykola-mokhnach: not sure if these tests are anyhow useful

harsha509:
Hi @mykola-mokhnach,

The IPA file is not removed automatically after install or uninstall. Cleanup must be handled via the AFC service.

Currently, the installation service only installs IPA assumes that the AFC service has already pushed the IPA into `PublicStaging` (tests does the same, it uses both **Afc + Installation**  service), and it only handles install/uninstall commands.

Should we create a utility helper that combines the AFC service with the installation service to properly handle both pushing and removing the IPA.

Please suggest!

src/services/ios/installation-proxy/index.ts
```diff
@@ -0,0 +1,646 @@
...
+    };
+
+    const conn = await this.getConnection();
+    conn.sendPlist(request);
+
+    const result: AppInfo[] = [];
+    const startTime = Date.now();
+
+    while (true) {
+      if (Date.now() - startTime > MAX_BROWSE_DURATION_MS) {
```
mykola-mokhnach: consider using performance.now() instead to measure time intervals

src/services/ios/installation-proxy/index.ts
```diff
@@ -0,0 +1,646 @@
...
+    try {
+      if (this.connection) {
+        this.connection.close();
+        this.connection = null;
+        log.debug('Connection closed successfully');
+      }
+    } catch (error) {
+      log.error('Error closing connection:', error);
+      // Always set to null even if close fails
+      this.connection = null;
```
mykola-mokhnach: this could be moved to a finally section to avoid repetition

src/services/ios/installation-proxy/index.ts
```diff
@@ -0,0 +1,646 @@
...
+        log.debug('Operation complete');
+        break;
+      }
+    }
+  }
+
+  private checkForError(
+    response: ProgressResponse | BrowseResponse | LookupResponse,
+  ): void {
+    if (response.Error) {
```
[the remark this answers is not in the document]
harsha509: updated with commit https://github.com/appium/appium-ios-remotexpc/pull/126/commits/3ffe81df44bed9676145e35ca4662d040daa28ee

src/services/ios/installation-proxy/index.ts
```diff
@@ -0,0 +1,646 @@
...
+    }
+
+    const result = await this.lookup(bundleIds, options);
+
+    if (calculateSizes) {
+      const sizeOptions: LookupOptions = {
+        ...options,
+        returnAttributes: SIZE_ATTRIBUTES,
+      };
+      const additionalInfo = await this.lookup(bundleIds, sizeOptions);
```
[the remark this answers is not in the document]
harsha509: updated with commit https://github.com/appium/appium-ios-remotexpc/pull/126/commits/3ffe81df44bed9676145e35ca4662d040daa28ee

PR #128, opened 2026-01-06
feat(afc): invoke callback for dirs too

navin772:
Invoke callback when creating directories, currently it is called only for files.

Useful for logging the no. of folders pulled in this xcuitest PR - https://github.com/appium/appium-xcuitest-driver/pull/2696

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## [0.24.3](https://github.com/appium/appium-ios-remotexpc/compare/v0.24.2...v0.24.3) (2026-01-07)
+
+### Miscellaneous Chores
+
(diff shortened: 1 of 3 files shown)


src/services/ios/afc/index.ts
```diff
@@ -627,11 +627,17 @@ export class AfcService {
       if (!dirCreated) {
         await fsp.mkdir(localDirPath, { recursive: true });
         dirCreated = true;
+        if (callback) {
+          await callback(remoteSrcDir, localDirPath);
```
mykola-mokhnach: how do we distinguish files from folders in the callback?

mykola-mokhnach: this would add an unnecessary performance hit. I would rather include this information into callback arguments

navin772: Good point, that would double the no. of calls, I will add a `isDirectory` param to the callback.

Issue #133, opened 2026-01-18
Fix: AFC file upload race condition causing intermittent INVALID_ARG errors


harsha509:
## Problem

The AFC + Installation Proxy integration tests fail intermittently (approximately 1 in 5 runs) with the following errors:

1. `AFC operation FILE_CLOSE failed with INVALID_ARG (7)`
2. `MissingPackagePath: No description` during installation

## Steps to Reproduce

Run the integration test multiple times:

```bash
for i in {1..10}; do
  echo "Run $i"
  npm run test:installation-workflow
done
```

Expected: All runs should pass
Actual: ~20% of runs fail with the errors above

## Root Cause

Race condition in [`AfcService.writeFromStream`](../../../src/services/ios/afc/index.ts#L329-L342):

```typescript
async writeFromStream(filePath: string, stream: Readable): Promise<void> {
  const handle = await this.fopen(filePath, 'w');
  const writeStream = this.createWriteStream(handle);
  try {
    await pipeline(stream, writeStream);
    log.debug(`Successfully wrote file: ${filePath}`);
  } catch (error) {
    await this.rmSingle(filePath, true);
    throw error;
  } finally {
    await this.fclose(handle);  // ⚠️ Race condition here
  }
}
```

**The issue**: After `pipeline()` completes, Node.js has finished writing to the AFC stream, but iOS may still be buffering/flushing data to disk. When `fclose()` is called immediately:
- iOS sometimes returns `INVALID_ARG` because file operations aren't complete
- Even if close succeeds, the file may not be fully available for installation

## Proposed Solution

Add a small delay (~150ms) after the pipeline completes to allow iOS to flush file buffers:

```typescript
async writeFromStream(filePath: string, stream: Readable): Promise<void> {
  const handle = await this.fopen(filePath, 'w');
  const writeStream = this.createWriteStream(handle);
  try {
    await pipeline(stream, writeStream);
    // Wait for iOS to flush buffers
    await new Promise(resolve => setTimeout(resolve, 150));
    log.debug(`Successfully wrote file: ${filePath}`);
  } catch (error) {
    await this.rmSingle(filePath, true);
    throw error;
  } finally {
    await this.fclose(handle);
  }
}
```

Apply the same fix to `setFileContents()` for consistency.

## Impact

- **Performance**: Adds ~150ms per file upload (negligible)
- **Reliability**: Eliminates intermittent test failures
- **User Experience**: More stable file operations

## Related Files

[...]

PR #135, opened 2026-01-20
fix: AFC file upload race condition

harsha509:
# AFC File Upload Race Condition - Research & Solution

## Problem

AFC + Installation Proxy integration tests fail intermittently (~20% of runs):
- `AFC operation FILE_CLOSE failed with INVALID_ARG (7)`
- `MissingPackagePath: No description` during installation

## Root Cause

iOS buffers AFC write operations in memory before committing to NAND flash. When `fclose()` is called immediately after write completes, the file may not be fully flushed to disk.

### Evidence

> "the device's operating system (iOS) **may buffer these writes in memory** ("in-core state") before writing them to the physical NAND flash."

Source: [libimobiledevice AFC documentation](https://docs.libimobiledevice.org/libimobiledevice/latest/afc_8h.html)

> "If files appear **truncated or missing after an AFC transfer**, it is often due to the thread finishing **before the device has committed the buffers**."

## Solution: FILE_LOCK Operation

The AFC protocol includes `FILE_LOCK` (opcode `0x0000001B`) for file synchronization.

### Standard Pattern from libimobiledevice

```c
afc_file_open(client, path, AFC_FOPEN_RW, &handle);
afc_file_lock(client, handle, AFC_LOCK_EX);  // Lock before write
afc_file_write(client, handle, data, len, &written);
afc_file_lock(client, handle, AFC_LOCK_UN);  // Unlock (triggers flush)
afc_file_close(client, handle);
```

### Lock Constants from pymobiledevice3

```python
AFC_LOCK_SH = 1 | 4   # 5: Shared lock (multiple readers)
AFC_LOCK_EX = 2 | 4   # 6: Exclusive lock (single writer)
AFC_LOCK_UN = 8 | 4   # 12: Unlock (triggers buffer flush)
```

### Why It Works

The UNLOCK operation triggers iOS to flush buffers to disk - similar to how POSIX file locks work on Unix systems where unlocking forces an fsync.

> "**`fsync(int fd)`**: Moves all modified data and attributes of a file descriptor to the permanent storage device.
>
> **`F_FULLFSYNC`**: A specific `fcntl` command on iOS/macOS that requests the drive to **flush its internal buffers to the physical flash cells**."

Source: [Apple Developer - fsync(2) man page](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/fsync.2.html)

The AFC daemon (`afcd`) on iOS uses these primitives. File lock release triggers similar synchronization.

## References

- [libimobiledevice AFC API Documentation](https://docs.libimobiledevice.org/libimobiledevice/latest/afc_8h.html)
- [The Apple Wiki - AFC Protocol Operations](https://theapplewiki.com/wiki/AFC)
- [newosxbook.com - AFCDecomposed (afcd daemon analysis)](http://newosxbook.com/articles/afcdecomposed.html)
- [Apple Developer - fsync(2) Documentation](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/fsync.2.html)


harsha509:
Fixes: #133

PR #136, opened 2026-01-21
feat(dvt): add ProcessControl instrument for process management

harsha509:


src/lib/types.ts
```diff
@@ -934,6 +939,51 @@ export interface DeviceInfoService {
   nameForGid(gid: number): Promise<string>;
 }
 
+/**
+ * ProcessControl service interface for managing processes
+ */
+export interface ProcessControlService {
+  /**
+   * Send a signal to a process
```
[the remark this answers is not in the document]
mykola-mokhnach: same below

src/lib/types.ts
```diff
@@ -934,6 +939,51 @@ export interface DeviceInfoService {
...
+  signal(pid: number, sig: number): Promise<any>;
+
+  /**
+   * Terminate a process
+   * @param pid The process identifier to kill
+   */
+  kill(pid: number): Promise<void>;
+
+  /**
+   * Disable memory limits for a specific process
```
mykola-mokhnach: how does this work? what are the default memory limits?

src/lib/types.ts
```diff
@@ -934,6 +939,51 @@ export interface DeviceInfoService {
...
+   * @param pid The process identifier
+   */
+  disableMemoryLimitForPid(pid: number): Promise<void>;
+
+  /**
+   * Get the process identifier for a bundle identifier
+   * @param bundleId The bundle identifier
+   * @returns The process identifier (PID)
+   */
+  processIdentifierForBundleIdentifier(bundleId: string): Promise<number>;
```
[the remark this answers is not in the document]
mykola-mokhnach: consider including a verb to the method name

src/services/ios/dvt/nskeyedarchiver-encoder.ts
```diff
@@ -0,0 +1,146 @@
+import { PlistUID } from '../../../lib/plist/index.js';
+
+/**
+ * Encodes JavaScript objects into NSKeyedArchiver format
+ * capable of satisfying NSSecureCoding requirements.
+ */
+export class NSKeyedArchiverEncoder {
```
mykola-mokhnach: Is this class covered by unit tests?

test/integration/process-control-test.ts
```diff
@@ -0,0 +1,105 @@
+import { logger } from '@appium/support';
+import { expect } from 'chai';
+
+import type { DVTServiceWithConnection } from '../../src/index.js';
+import * as Services from '../../src/services.js';
+
+const log = logger.getLogger('ProcessControl.test');
+
+describe('ProcessControl Service', function () {
```
navin772: It would be nice to have some tests for `outputEvents` generator, we can add steps/requirements as comments in tests if required.

src/services/ios/dvt/instruments/process-control.ts
```diff
@@ -0,0 +1,216 @@
...
+        if (selector === 'outputReceived:fromProcess:atTime:') {
+          // Auxiliaries format: [message (string), pid (int), timestamp (obj/long)]
+
+          if (auxiliaries && auxiliaries.length >= 2) {
+            const message = auxiliaries[0];
+            const pid = auxiliaries[1];
+            const timestampRaw = auxiliaries[2];
+
+            let timestamp: bigint | undefined;
+            let date: Date | undefined;
```
navin772: We never populate `date` after this, so it will return `undefined`. Need to convert and assign.

src/lib/types.ts
```diff
@@ -934,6 +939,51 @@ export interface DeviceInfoService {
...
+   * Terminate a process
+   * @param pid The process identifier to kill
+   */
+  kill(pid: number): Promise<void>;
+
+  /**
+   * Disable memory limits for a specific process
+   * @param pid The process identifier
+   */
+  disableMemoryLimitForPid(pid: number): Promise<void>;
```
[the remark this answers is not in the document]
harsha509: HI Mykola,

calling `disableMemoryLimitForPid` twice, the device returns true both times and the second call and nothing breaks. The second call has no effect since the limit is already disabled

Added a note in the JSDoc documenting this idempotent behavior and added a test. commit https://github.com/appium/appium-ios-remotexpc/pull/136/commits/b38bad2cb4ff7b72e53bf8bec54fdf7188758a10

src/services/ios/dvt/nskeyedarchiver-encoder.ts
harsha509: Updated with commit https://github.com/appium/appium-ios-remotexpc/pull/136/commits/cb9f73ee0135bf8fd01929f654d95daa838089f4

src/services/ios/dvt/nskeyedarchiver-encoder.ts
```diff
@@ -0,0 +1,146 @@
...
+    if (Buffer.isBuffer(value)) {
+      index = this.archiveBuffer(value);
+    } else if (Array.isArray(value)) {
+      index = this.archiveArray(value);
+    } else if (typeof value === 'object') {
+      // Treat generic objects as dictionaries
+      index = this.archiveDictionary(value);
+    } else {
+      // Fallback (e.g. symbols, functions)
+      throw new Error(`Unsupported type for NSKeyedArchiver: ${typeof value}`);
```
mykola-mokhnach: is it safe to throw errors in this method? AFAIK encode and decode are invoked in pipelines, where uncontrolled errors may break flows with unexpected consequences

mykola-mokhnach (approved):

PR #137, opened 2026-01-30
feat(installation-proxy): make operation timeout configurable

navin772:
Makes timeouts for install/upgrade/uninstall configurable as discussed in https://github.com/appium/appium-xcuitest-driver/pull/2714#discussion_r2742914905

Defaults to the `MAX_INSTALL_DURATION_MS` const (10 mins) if no `operationTimeout` is passed.

### DIFF
--- a/.github/workflows/format-check.yml
+++ b/.github/workflows/format-check.yml
@@ -7,36 +7,21 @@ on:
     branches: [ main ]
 
 jobs:
-  prepare_matrix:
-    runs-on: ubuntu-latest
-    outputs:
-      versions: ${{ steps.generate-matrix.outputs.lts }}
-    steps:
-    - name: Select all current LTS versions of Node.js
-      id: generate-matrix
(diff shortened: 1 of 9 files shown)


src/services/ios/installation-proxy/index.ts
```diff
@@ -193,11 +193,13 @@ export class InstallationProxyService extends BaseService {
    * @param packagePath Path to the IPA file on the device (e.g., '/PublicStaging/app.ipa')
    * @param options Installation options
    * @param progressCallback Optional callback for progress updates
+   * @param operationTimeout Optional timeout in milliseconds for the entire operation (default: 10 minutes)
    */
   async install(
     packagePath: string,
     options: InstallOptions = {},
     progressCallback?: ProgressCallback,
+    operationTimeout?: number,
```
mykola-mokhnach: method signatures are getting confusing if they contain more than 3 arguments. Consider moving this argument under install options. same below

PR #147, opened 2026-02-27
feat: add testmanagerd RSD service

navin772:
For testing, test against a sample app with valid XCTest Bundle using the command:

```
UDID=<device-udid> TEST_RUNNER_BUNDLE_ID=<xctrunner-bundle-id> APP_UNDER_TEST_BUNDLE_ID=<target-app-bundle-id> XCTEST_BUNDLE_ID=<xctest-bundle-id> npm run test:testmanagerd
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,15 @@
+## [0.30.0](https://github.com/appium/appium-ios-remotexpc/compare/v0.29.1...v0.30.0) (2026-02-16)
+
+### Features
+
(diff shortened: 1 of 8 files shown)


src/services/ios/dvt/nskeyedarchiver-encoder.ts
```diff
@@ -62,8 +62,15 @@ export class NSKeyedArchiverEncoder {
     } else if (Array.isArray(value)) {
       index = this.archiveArray(value);
     } else if (typeof value === 'object') {
-      // Treat generic objects as dictionaries
-      index = this.archiveDictionary(value);
+      // Check for marker types used by testmanagerd protocol
+      if (value.__type === 'NSUUID') {
+        index = this.archiveNSUUID(value.uuid);
```
mykola-mokhnach: I assume this class is generic and it should not know about xctest-specific primitives. Consider inheriting it instead

PR #148, opened 2026-03-02
feat: add testmanagerd DTX service for XCTest

navin772:
Implement `DvtTestmanagedProxyService, a DTX protocol client for `com.apple.dt.testmanagerd.remote`.

This enables `testmanagerd` channel creation, encoding and message exchange for XCTest sessions.

PR #151, opened 2026-03-06
feat(testmanagerd): add XCTestConfiguration encoder for XCTest session setup

navin772:
Adds XCTestConfigurationEncoder extending `TestmanagerdEncoder` with support for `NSURL` and XCTestConfiguration NSKeyedArchiver encoding. 

This is the serialization layer needed to send test configuration to the device during XCTest session initialization.

Follow up PR for #148


navin772:
Hi @mykola-mokhnach, could you review this when you get time

PR #157, opened 2026-03-14
chore: remove no endpoint log

KazuCocoa:
I assume this endpoint doesn't work

https://github.com/appium/appium-ios-remotexpc/blob/main/src/lib/tunnel/tunnel-registry-server.ts#L179

### DIFF
--- a/scripts/start-appletv-tunnel.ts
+++ b/scripts/start-appletv-tunnel.ts
@@ -160,7 +160,6 @@ async function main(): Promise<void> {
     log.info(
       `   - GET /remotexpc/tunnels/${deviceInfo.identifier} - Get tunnel by identifier`,
     );
-    log.info('   - GET /remotexpc/tunnels/metadata - Get registry metadata');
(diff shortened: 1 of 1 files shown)


scripts/start-appletv-tunnel.ts
```diff
@@ -160,7 +160,6 @@ async function main(): Promise<void> {
     log.info(
       `   - GET /remotexpc/tunnels/${deviceInfo.identifier} - Get tunnel by identifier`,
     );
-    log.info('   - GET /remotexpc/tunnels/metadata - Get registry metadata');
 
     log.info('\n✅ Apple TV tunnel is ready!');
     log.info(`Use: --rsd ${tunnel.Address} ${tunnel.RsdPort}`);
```
KazuCocoa: Not directly related to this PR, but what is this argument? How to use this `--rsd`? I assume for apple tv, we need to run this after this tunnel creation

PR #189, opened 2026-04-12
fix: Make sure wired devices are listed before wireless ones

mykola-mokhnach:


src/lib/usbmux/utils.ts
```diff
@@ -0,0 +1,59 @@
...
+      return a.indexInSlice - b.indexInSlice;
+    })
+    .map(({ item }) => item);
+}
+
+/** Wireless usbmux rows (same UDID as USB) use ConnectionType `Network`. */
+function wirelessRank<T extends { Properties: { ConnectionType: string } }>(
+  device: T,
+): number {
+  return device.Properties.ConnectionType === 'Network' ? 1 : 0;
```
KazuCocoa, line 57: Btw, could we add type for ConnectionType with Network or USB (instead of string)?

KazuCocoa (approved):
lg entirely

src/lib/usbmux/utils.ts
mykola-mokhnach: this would be a breaking change. also, it might become confusing if Apple eventually decides to add some other value there.

src/lib/usbmux/utils.ts
```diff
@@ -0,0 +1,59 @@
+/**
+ * For duplicate UDIDs (USB + Wi‑Fi), order wired/USB before `Network` entries.
+ * Reorders **in place across the original indices** for each UDID (no pulling one
+ * UDID into a contiguous block). Among equal rank, input order is kept (stable).
```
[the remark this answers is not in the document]
mykola-mokhnach: fixed

PR #190, opened 2026-04-21
refactor: remove deprecated discovery and WebInspector APIs

mykola-mokhnach:
BREAKING CHANGE: DiscoveredDeviceMetadata no longer includes minVersion, authTag, or serviceType.
BREAKING CHANGE: AppleTVDevice no longer includes minVersion, authTag, or interfaceIndex.
BREAKING CHANGE: WebInspectorService.stopListening() removed; use stopListeningAsync() instead.

PR #191, opened 2026-04-21
fix: Revert "refactor: remove deprecated discovery and WebInspector APIs"

mykola-mokhnach:
Reverts appium/appium-ios-remotexpc#190


mykola-mokhnach:
it looks like versioning script needs some fixes to be able to publish v1.0.0

PR #199, opened 2026-05-05
fix: Tune emitted events on the port forwarder

mykola-mokhnach:


src/lib/port-forwarding/types.ts
```diff
@@ -37,9 +37,10 @@ export interface DevicePortForwarderEvents {
   started: () => void;
   stopped: () => void;
   clientConnected: (socket: Socket) => void;
-  clientDisconnected: (socket: Socket) => void;
+  clientDisconnected: (socket: Socket, error?: Error) => void;
   upstreamConnected: (socket: Socket) => void;
-  upstreamDisconnected: (socket: Socket) => void;
+  upstreamDisconnected: (socket: Socket, error?: Error) => void;
   upstreamConnectError: (error: unknown) => void;
+  /** @deprecated Not used, will be removed in the next major version. */
```
mykola-mokhnach, line 44: the reason for deprecating this is the default node behaviour which treats unhandled error events as unhandled exceptions

src/lib/port-forwarding/device-port-forwarder.ts
```diff
@@ -107,13 +107,9 @@ export class DevicePortForwarder extends EventEmitter {
     ) as this;
   }
 
-  private async handleLocalConnection(localSocket: Socket): Promise<void> {
+  private async handleDownstreamConnection(localSocket: Socket): Promise<void> {
     this.activeSockets.add(localSocket);
     this.emit('clientConnected', localSocket);
-    localSocket.once('close', () => {
```
harsha509, line 113: Removing this listener may leak activeSockets on every failed upstream connect.
Should we delete socket somewhere in catch?

mykola-mokhnach: I didn't remove it, just moved below. Sockets are still deleted properly on teardown

mykola-mokhnach: Actually found one missing deletion in the `catch` section below. Should be ok now. Good catch btw 👍

KazuCocoa (approved):

harsha509 (approved):

PR #202, opened 2026-05-06
fix(packet-stream-server): only evict client on actual broadcast write error

harsha509:
## Summary

The old code passed an "error handler" to socket.write, but Node calls that handler on every write ,even successful ones (with no error). So every successful broadcast was logged as a failure and the client was wrongly removed. After the first packet, the client list was empty and no one received anything.

The fix: only log and remove the client when there's an actual error.

## Changes

- Skip destroyed clients up front via `continue` (small readability win).
- Pass an `(err?: Error | null) => void` callback to `client.write` and only log + evict the client when `err` is truthy.

### DIFF
--- a/src/lib/tunnel/packet-stream-server.ts
+++ b/src/lib/tunnel/packet-stream-server.ts
@@ -86,7 +86,7 @@ export class PacketStreamServer extends EventEmitter {
   private createPacketConsumer(): PacketConsumer {
     return {
       onPacket: (packet: PacketData) => {
-        this.broadcastPacket(packet);
+        void this.broadcastPacket(packet);
       },
     };
   }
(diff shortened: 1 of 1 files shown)


src/lib/tunnel/packet-stream-server.ts
```diff
@@ -86,33 +86,48 @@ export class PacketStreamServer extends EventEmitter {
...
       log.error(`Failed to broadcast packet: ${err}`);
     }
   }
 
+  /**
+   * Promisified wrapper around `socket.write` so callers can use async/await.
+   */
+  private writeToClient(client: Socket, message: Buffer): Promise<void> {
+    return new Promise<void>((resolve, reject) => {
+      client.write(message, (err?: Error | null) =>
```
mykola-mokhnach: this generates a linter warning, lets move the callback to a separate variable avoid it

src/lib/tunnel/packet-stream-server.ts
```diff
@@ -86,33 +86,48 @@ export class PacketStreamServer extends EventEmitter {
...
-          this.clients.delete(client);
-        };
-        if (!client.destroyed) {
-          client.write(message, onError);
-        }
-      }
+      await Promise.all(
+        Array.from(this.clients).map(async (client) => {
+          if (client.destroyed) {
+            return;
```
mykola-mokhnach, line 106: should we also delete the client from clients list in such case?

src/lib/tunnel/packet-stream-server.ts
```diff
@@ -86,33 +86,48 @@ export class PacketStreamServer extends EventEmitter {
...
-          log.error(`Failed to write to client: ${err}`);
-          this.clients.delete(client);
-        };
-        if (!client.destroyed) {
-          client.write(message, onError);
-        }
-      }
+      await Promise.all(
+        Array.from(this.clients).map(async (client) => {
+          if (client.destroyed) {
```
mykola-mokhnach: should we also check if the socket is writeable?

harsha509: agreed, updated with commit https://github.com/appium/appium-ios-remotexpc/pull/202/changes/f9358716932ed290924bbd7f70633b37958e0bd9

src/lib/tunnel/packet-stream-server.ts
harsha509: agreed, updated with commit https://github.com/appium/appium-ios-remotexpc/pull/202/changes/f9358716932ed290924bbd7f70633b37958e0bd9

src/lib/tunnel/packet-stream-server.ts
harsha509: updated with commit https://github.com/appium/appium-ios-remotexpc/pull/202/changes/bd825e7a7531bceb0f1bf706ae6dd70179e7eb4f

mykola-mokhnach (approved):

KazuCocoa (approved):

PR #205, opened 2026-05-11
fix(apple-tv): correct mDNS service usage for pair-setup and tunnel

### LINKED ISSUE #204: pair-appletv fails with AppleTV 26.4 (wi-fi)
I am unable to pair AppleTV 26.4 with the following error:

```
 % APPIUM_IOS_REMOTEXPC_LOG_LEVEL=silly appium driver run xcuitest pair-appletv                                
info AppleTVPairingService Discovering Apple TV devices (waiting 10 seconds)...
dbug Devicectl Executing "xcrun" "devicectl" "list" "devices" "--quiet" "--json-output" "-"
info AppleTVPairingService Found 1 device(s):
info AppleTVPairingService   [0] 86585DAF-DA7C-4A0C-A0A4-4B9F8E0ADA76 (00008110-001160343E11A01E) - AppleTV14,1 v26.4
Press Enter to select device [0], or enter index: 
info AppleTVPairingService Selected device: 86585DAF-DA7C-4A0C-A0A4-4B9F8E0ADA76 (00008110-001160343E11A01E)
info AppleTVPairingService Connecting to device 86585DAF-DA7C-4A0C-A0A4-4B9F8E0ADA76 at 192.168.8.149:49152
dbug NetworkClient Connecting to 192.168.8.149:49152
dbug NetworkClient Connected successfully
dbug NetworkClient Sending packet: { size: 188 }
dbug NetworkClient Response header parsed: expecting 397 bytes
dbug NetworkClient Response received and parsed successfully
dbug PairingProtocol Handshake completed
dbug NetworkClient Sending packet: { size: 236 }
dbug NetworkClient Response header parsed: expecting 389 bytes
dbug NetworkClient Response received and parsed successfully
dbug NetworkClient Sending packet: { size: 121 }
dbug PairingProtocol Pair verification attempt completed
dbug NetworkClient Sending packet: { size: 222 }
Warning: Detected unsettled top-level await at file:///Users/device-lab/.appium/node_modules/appium-xcuitest-driver/scripts/pair-appletv.mjs:56
await main();
^



Error: ✖ Encountered an error when running 'pair-appletv': Script exited with code 13
```

Apple TV 4K (3rd generation)
tvOS: 26.4
Xcode: 26.4
Macos: 26.4.1
Appium 3.3.1
appium-xcuitest-driver 10.36.4 (I'm aware it's not the newest but 11.2.1 behaves exactly the same)
appium-ios-remotexpc 0.42.1
npm 10.9.7
node 22.22.2

Apple TV is paired with Xcode, WDA is signed. 
I followed [these steps](https://github.com/appium/appium-ios-remotexpc/blob/main/docs/apple-tv-pairing-guide.md) but I never reach the PIN viewing on the Apple TV screen.

I will appreciate any suggestions what I am missing and how to make this work to complete the pairing process.

harsha509:
PR #171 collapsed the pair-setup and tunnel-creation flows onto a single discovery service type, silently regressing both flows:

- Pair-setup must use _remotepairing-manual-pairing._tcp (the PIN-mode service); it was incorrectly switched to _remotepairing._tcp, where PIN setup does not work.
- Tunnel createListener must use _remotepairing._tcp (the trusted/ verified service that appears once a host has paired); requesting it on the manual-pairing service fails with "Tunnel listener creator not set".

Restoring _remotepairing-manual-pairing._tcp for pair-setup also exposed that the dnssd npm library can't see Apple's long service names (RFC 6335 forbids >15 chars). On darwin we now wrap the system dns-sd CLI (BonjourDiscoveryBackend) which has no such restriction. Other platforms continue to use the dnssd library.

Tunnel handshake additionally fails on Macs where remoted already holds the trusted tunnel for the device (the device refuses a second createListener). RemotedController SIGSTOPs remoted for the duration of the handshake and SIGCONTs it in a finally block

Other clean-ups in the same patch:

- Introduce BaseDiscoveryBackend to share the in-flight deduplication and option defaulting between the dnssd and bonjour backends.
- Extract normalizeHostname / resolveIpAddress into discovery-utils.
- Drop the misleading REMOTE_PAIRING_DISCOVERY_SERVICE_TYPE alias and make every caller import the explicit _MANUAL_ or _VERIFIED_ constant.
- Harden RemotedController: pgrep -x remoted (robust against launchd arguments) and an actionable hint when SIGSTOP fails with EPERM.

fixes : https://github.com/appium/appium-ios-remotexpc/issues/204


src/lib/apple-tv/tunnel/remoted-controller.ts
```diff
@@ -0,0 +1,89 @@
...
+ * behavior and is a no-op on non-darwin platforms.
+ */
+export class RemotedController {
+  private suspendedPid: number | null = null;
+
+  /**
+   * Suspends `remoted` if it is running and not already suspended.
+   * Safe to call multiple times.
+   */
+  async suspendIfRequired(): Promise<void> {
```
mykola-mokhnach: not sure about the method name. What `ifRequired` means in this context?

src/lib/apple-tv/tunnel/remoted-controller.ts
```diff
@@ -0,0 +1,89 @@
...
+}
+
+/**
+ * Locate the running `remoted` process by exact process name.
+ * Uses `pgrep -x` (exact match on process name) which is robust against
+ * launchd adding arguments to the daemon's command line.
+ */
+function findRemotedPid(): Promise<number | null> {
+  return new Promise((resolve) => {
+    const child = spawn('pgrep', ['-x', REMOTED_PROCESS_NAME]);
```
mykola-mokhnach: consider using execFileAsync:

```
import { execFile } from 'node:child_process';
import { promisify } from 'node:util';

const execFileAsync = promisify(execFile);

async function findPidsByName(name) {
  try {
    const { stdout } = await execFileAsync('pgrep', ['-x', name]);

    return stdout
      .trim()
      .split('\n')
      .filter(Boolean)
      .map(Number);
  } catch (err) {
    // pgrep exits with code 1 when no processes match
    if (err.code === 1) {
      return [];
    }

    throw err;
  }
}

const pids = await findPidsByName('Finder');
console.log(pids);
```

src/lib/discovery/bonjour-discovery-backend.ts
```diff
@@ -0,0 +1,300 @@
...
+  DiscoveredDeviceMetadata,
+  DiscoveryOptions,
+} from './types.js';
+
+const log = getLogger('BonjourDiscoveryBackend');
+
+const DNS_SD_BIN = 'dns-sd';
+const RESOLVE_TIMEOUT_MS = 3000;
+
+const BROWSE_LINE =
```
mykola-mokhnach: this regexp could be fragile. 
Consider instead just splitting the line by \s+ and then taking the value of the last column

src/lib/discovery/bonjour-discovery-backend.ts
```diff
@@ -0,0 +1,300 @@
...
+  /^\s*\d{2}:\d{2}:\d{2}\.\d+\s+(Add|Rmv)\s+\d+\s+\d+\s+(\S+)\s+(\S+)\s+(.+)$/;
+const REACHABLE = /can be reached at (\S+):(\d+)/;
+const TXT_PAIR = /(\S+?)=([^\s]+)/g;
```
mykola-mokhnach: `\S is the same as [^\s]`

src/lib/discovery/bonjour-discovery-backend.ts
```diff
@@ -0,0 +1,300 @@
...
+
+/**
+ * Parse one `dns-sd -B` output line into an action + service descriptor.
+ *
+ * Example matched line:
+ *   "11:52:30.137  Add        2  17 local.   _remotepairing-manual-pairing._tcp. Bedroom"
+ */
+function parseBrowseLine(
+  line: string,
+): { action: 'Add' | 'Rmv'; service: BrowsedService } | null {
```
mykola-mokhnach: consider adding a separate typedef for the return type

src/lib/discovery/bonjour-discovery-backend.ts
```diff
@@ -0,0 +1,300 @@
...
+/**
+ * Parse the space-separated `key=value` pairs that follow a `-L` reachable line.
+ */
+function parseTxtRecord(line: string): Record<string, string> {
+  const txt: Record<string, string> = {};
+  let match: RegExpExecArray | null;
+  while ((match = TXT_PAIR.exec(line)) !== null) {
+    txt[match[1]] = match[2];
+  }
+  TXT_PAIR.lastIndex = 0;
```
mykola-mokhnach: we must not mutate constants

src/lib/discovery/discovery-backend-factory.ts
```diff
@@ -1,13 +1,21 @@
+import { BonjourDiscoveryBackend } from './bonjour-discovery-backend.js';
 import { DnssdDiscoveryBackend } from './dnssd-discovery-backend.js';
 import type { DiscoveryOptions, IDeviceDiscoveryBackend } from './types.js';
 
 /**
  * Create the default device discovery backend for the current platform.
+ *
+ * On macOS we prefer the system `dns-sd` CLI (Bonjour) because Apple's
+ * mDNSResponder advertises non-RFC-6335-compliant service names (e.g.
```
mykola-mokhnach, line 9: is this is the case then probably dnssd won't work on linux as well 🤷 , since the device advertises itself in a non-standard manner

src/lib/discovery/dnssd-discovery-backend.ts
```diff
@@ -1,51 +1,34 @@
...
+    super(options);
   }
 
-  private async runDiscovery(timeoutMs: number): Promise<DiscoveredDevice[]> {
-    const serviceType =
-      this.options.serviceType || DISCOVERY_DEFAULT_SERVICE_TYPE;
-    const domain = this.options.domain || DISCOVERY_DEFAULT_DOMAIN;
-    const browser = new dnssd.Browser(serviceType, {
-      domain,
+  protected async runDiscovery(timeoutMs: number): Promise<DiscoveredDevice[]> {
```
mykola-mokhnach: override?

test/unit/apple-tv/remoted-controller.spec.ts
```diff
@@ -0,0 +1,177 @@
+import { expect } from 'chai';
```
mykola-mokhnach: not sure if these all tests are useful. I would delete them

src/lib/discovery/bonjour-discovery-backend.ts
```diff
@@ -0,0 +1,300 @@
...
+ * Build a user-friendly error for `dns-sd -B` failures.
+ */
+function buildBrowseError(
+  spawnError: NodeJS.ErrnoException,
+  stderr: string,
+): Error {
+  const detail = stderr.trim() || spawnError.message;
+  const message =
+    spawnError.code === 'ENOENT'
+      ? `\`dns-sd\` binary not found in PATH: ${detail}`
```
mykola-mokhnach: Please use the constant for the binary name

src/lib/discovery/bonjour-discovery-backend.ts
```diff
@@ -0,0 +1,300 @@
...
+/**
+ * Build a user-friendly error for `dns-sd -B` failures.
+ */
+function buildBrowseError(
+  spawnError: NodeJS.ErrnoException,
+  stderr: string,
+): Error {
+  const detail = stderr.trim() || spawnError.message;
+  const message =
+    spawnError.code === 'ENOENT'
```
mykola-mokhnach, line 178: we could use fs.which to verify the binary is present in PATH

src/lib/discovery/bonjour-discovery-backend.ts
```diff
@@ -0,0 +1,300 @@
...
+/**
+ * Run `dns-sd -B` for the configured timeout and return all currently-present
+ * services (Add not followed by Rmv).
+ *
+ * Rejects when the `dns-sd` binary cannot be spawned (e.g. missing from PATH)
+ * or when it terminates with a non-zero exit code before we time it out, so
+ * callers can distinguish a broken discovery transport from "no devices
+ * found".
+ */
+function browseServices(
```
mykola-mokhnach: async function

src/lib/discovery/bonjour-discovery-backend.ts
```diff
@@ -0,0 +1,300 @@
...
+    port: resolved.port,
+    metadata,
+  };
+}
+
+/**
+ * Run `dns-sd -L` and return as soon as we have a reachable line + TXT line,
+ * or after `timeoutMs`.
+ */
+function runResolve(
```
mykola-mokhnach: async function

src/lib/discovery/bonjour-discovery-backend.ts
```diff
@@ -0,0 +1,300 @@
+import { type ChildProcess, spawn } from 'node:child_process';
```
mykola-mokhnach, line 2: it would be nice to cover output parsing helpers with unit tests

mykola-mokhnach:
Does dns-sd discovery run as expected if executed from sudo?

src/lib/apple-tv/tunnel/remoted-controller.ts
harsha509: `suspendIfRequired()`  to pause the macOS remoted background process so it doesn't fight us for the device tunnel, but only if needed 

Modified to just `suspend()`

src/lib/apple-tv/tunnel/remoted-controller.ts
harsha509: Replaced the manual spawn('pgrep', ...) + event-emitter wiring with promisify(execFile), exactly as suggested

src/lib/discovery/bonjour-discovery-backend.ts
harsha509: Replaced the fragile BROWSE_LINE regex with a `parseBrowseLine` helper that splits on \s+ and joins everything past the fixed columns as the name (preserves multi-word names like "Living Room TV").

src/lib/discovery/bonjour-discovery-backend.ts
harsha509: Replaced [^\s] with \S in TXT_PAIR.

src/lib/discovery/bonjour-discovery-backend.ts
harsha509: Added a `ParsedBrowseLine` typedef (with BrowseAction = 'Add' | 'Rmv') in bonjour-output-parsers.ts and used it as parseBrowseLine's return type.

src/lib/discovery/bonjour-discovery-backend.ts
harsha509: Switched to `String.prototype.matchAll(TXT_PAIR)` which doesn't touch the regex's lastIndex, no more constant mutation.

src/lib/discovery/discovery-backend-factory.ts
harsha509: YEs @mykola-mokhnach ,

dnssd enforces RFC 6335 length limits too, so on Linux/Windows it also can't see Apple's long names. Updated the factory doc to reflect that.

Tracking that as a follow-up: [#206](https://github.com/appium/appium-ios-remotexpc/issues/206).

src/lib/discovery/dnssd-discovery-backend.ts
harsha509: Added `override` modifier, runDiscovery is now protected override async. 
Same applied on `BonjourDiscoveryBackend` for consistency.

test/unit/apple-tv/remoted-controller.spec.ts
harsha509: Deleted test/unit/apple-tv/remoted-controller.spec.ts entirely.

src/lib/discovery/bonjour-discovery-backend.ts
harsha509: Replaced the hard-coded `dns-sd` strings in both error messages with the DNS_SD_BIN constant.

src/lib/discovery/bonjour-discovery-backend.ts
harsha509: Added an `assertDnsSdAvailable()` preflight at the top of runDiscovery that uses `fs.which` from `@appium/support` to verify `dns-sd` is on PATH before we spawn, surfaces a clear error otherwise.

src/lib/discovery/bonjour-discovery-backend.ts
harsha509: updated..!

src/lib/discovery/bonjour-discovery-backend.ts
harsha509: updated ...!

harsha509:
> Does dns-sd discovery run as expected if executed from sudo?

Yes Mykola,

`dns-sd` lives at `/usr/bin/dns-sd` which is in sudo's default `secure_path `on macOS, 
so it resolves the same under sudo. 

The new `assertDnsSdAvailable()` preflight uses `fs.which` so any unusual PATH stripping. surfaces a clear error before we try to spawn.

src/lib/discovery/bonjour-discovery-backend.ts
harsha509: updated

src/lib/apple-tv/tunnel/remoted-controller.ts
```diff
@@ -0,0 +1,100 @@
...
+      .split('\n')
+      .map((line) => parseInt(line.trim(), 10))
+      .filter((pid) => Number.isFinite(pid) && pid > 0);
+  } catch (err) {
+    const code = (err as { code?: unknown })?.code;
+    if (code === PGREP_NO_MATCH_EXIT_CODE) {
+      return [];
+    }
+    log.debug(`pgrep failed for "${name}": ${err}`);
+    return [];
```
mykola-mokhnach, line 97: since we anyway return an empty array the above condition could just decide whether to log or not

src/lib/discovery/bonjour-discovery-backend.ts
```diff
@@ -0,0 +1,273 @@
...
+ * Apple's Bonjour CLI installed).
+ */
+async function assertDnsSdAvailable(): Promise<void> {
+  try {
+    await fs.which(DNS_SD_BIN);
+  } catch (err) {
+    const cause = err instanceof Error ? err : new Error(String(err));
+    const message = `\`${DNS_SD_BIN}\` binary not found in PATH: ${cause.message}`;
+    log.warn(message);
+    throw new Error(message, { cause });
```
mykola-mokhnach: lint does not like this

src/lib/discovery/bonjour-discovery-backend.ts
```diff
@@ -0,0 +1,273 @@
...
+
+/**
+ * Run `dns-sd -L` and return as soon as we have a reachable line + TXT line,
+ * or after `timeoutMs`.
+ */
+async function runResolve(
+  svc: BrowsedService,
+  timeoutMs: number,
+): Promise<ResolvedService | null> {
+  return await new Promise<ResolvedService | null>((resolve) => {
```
mykola-mokhnach: this promise is very similar to the above one. Would it be possible to avoid unnecessary duplication?

src/lib/discovery/bonjour-output-parsers.ts
```diff
@@ -0,0 +1,102 @@
...
+  return { hostname: m[1], port };
+}
+
+/**
+ * Parse the space-separated `key=value` pairs that follow a `-L` reachable
+ * line into a TXT record map. Uses `String.prototype.matchAll` so we never
+ * mutate the shared regex's `lastIndex`.
+ */
+export function parseTxtRecord(line: string): Record<string, string> {
+  const txt: Record<string, string> = {};
```
mykola-mokhnach: this function could be simplified by using reduce on `line.matchAll(TXT_PAIR)`

src/lib/discovery/discovery-utils.ts
```diff
@@ -0,0 +1,35 @@
...
+  addresses?: string[],
+): Promise<string | undefined> {
+  if (addresses?.[0]) {
+    return addresses[0];
+  }
+  if (!host) {
+    return undefined;
+  }
+  try {
+    const results = await lookup(host.replace(/\.$/, ''), {
```
mykola-mokhnach: we can reuse the above stripTrailingDot helper

test/unit/discovery/bonjour-output-parsers.spec.ts
```diff
@@ -0,0 +1,122 @@
+import { expect } from 'chai';
```
mykola-mokhnach, line 1: 👍

mykola-mokhnach (approved):

src/lib/apple-tv/tunnel/remoted-controller.ts
harsha509: Done.
flipped the condition to gate just the log.debug(...), single return [] at the bottom of the catch.

src/lib/discovery/bonjour-discovery-backend.ts
harsha509: fixed...!

src/lib/discovery/bonjour-discovery-backend.ts
harsha509: Extracted a generic r`unDnsSd<T>(args, timeoutMs, handlers)` helper that owns the `spawn / readline / timer / cleanup lifecycle browseServices ` and `runResolve ` are now thin handler objects

src/lib/discovery/bonjour-output-parsers.ts
harsha509: parseTxtRecord now does Array.from(line.matchAll(TXT_PAIR)).reduce(...) into the record.

src/lib/discovery/discovery-utils.ts
harsha509: fixed..

PR #209, opened 2026-05-14
fix(services): close internal RemoteXpcConnection in start*Service helpers

harsha509:
`startAfcService`, `startSyslogService`, `startSyslogBinaryService` and `startSyslogTextService` each opened a RemoteXpcConnection for port discovery and never closed it. Combined with a caller that opens its own RSD, the leaked discovery RSD overlapped and `remoted` reset one of them, producing the `Connection error: read ECONNRESET` storm at
driver startup.

Wrap discovery + return in `try/finally { await remoteXPC.close() }`, mirroring the existing startXCTestServices pattern in the same file.

### DIFF
--- a/src/services.ts
+++ b/src/services.ts
@@ -201,8 +201,15 @@ export async function startPowerAssertionService(
 export async function startSyslogService(
   udid: string,
 ): Promise<SyslogServiceType> {
-  const { tunnelConnection } = await createRemoteXPCConnection(udid);
-  return new SyslogService([tunnelConnection.host, tunnelConnection.port]);
+  const { remoteXPC, tunnelConnection } = await createRemoteXPCConnection(udid);
+  try {
+    return new SyslogService([tunnelConnection.host, tunnelConnection.port]);
+  } finally {
+    // Release the discovery RSD eagerly so iOS's `remoted` daemon does not
+    // see this connection overlap with any subsequent RSD probe and reset
+    // it (the source of the ECONNRESET storm during driver startup).
+    await remoteXPC.close();
+  }
 }
 
 const RSD_SYSLOG_BINARY_SERVICE_NAME = 'com.apple.os_trace_relay.shim.remote';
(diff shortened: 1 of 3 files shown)


src/services.ts
```diff
@@ -201,8 +201,15 @@ export async function startPowerAssertionService(
 export async function startSyslogService(
   udid: string,
 ): Promise<SyslogServiceType> {
-  const { tunnelConnection } = await createRemoteXPCConnection(udid);
-  return new SyslogService([tunnelConnection.host, tunnelConnection.port]);
+  const { remoteXPC, tunnelConnection } = await createRemoteXPCConnection(udid);
```
mykola-mokhnach: would it make sense to have a `withRemoteXpcConnection` wrapper to avoid unnecessary duplication?

test/unit/services/start-services-cleanup.spec.ts
```diff
@@ -0,0 +1,200 @@
+import { expect } from 'chai';
```
mykola-mokhnach, line 1: not sure if these unit tests are useful

src/services.ts
harsha509: updated.

mykola-mokhnach:
are unit tests broken?

test/unit/services/start-services-cleanup.spec.ts
harsha509: removed redundant tests and kept what is needed

src/services.ts
```diff
@@ -214,6 +217,10 @@ export interface StartXCTestServicesOptions {
   includeInstallationProxy?: boolean;
 }
 
+type TunnelConnection = Awaited<
```
mykola-mokhnach: would it make sene to import type for `tunnelConnection` directly instead

harsha509:
> are unit tests broken?

https://github.com/appium/appium-ios-remotexpc/pull/210

mykola-mokhnach (approved):

src/services.ts
harsha509: Yes Mykola, updated with commit https://github.com/appium/appium-ios-remotexpc/pull/209/changes/0204d7e6f3e800fe57167d6f0f0d79b195cfd698 !

PR #214, opened 2026-05-18
fix: Tune remote xpc connect retry logic

mykola-mokhnach:
## Summary

Improves RemoteXPC reconnect behavior when the tunnel/RSD endpoint is not ready: failed attempts fail faster, retries stay within a bounded window, and instant failures no longer hammer the endpoint.

### Problem

`TunnelManager.createRemoteXPCConnection` could block for **~50+ seconds** on reconnect (15 retries × 3s timeout + 500ms delay between attempts), even when the socket failed immediately.

### Solution

**Timeouts (`remote-xpc-connection.ts`)**
- **500ms TCP connect timeout** (`CONNECTION_CONNECT_TIMEOUT_MS`) — separate from the per-attempt operation budget
- **Per-attempt operation timeout** via `connect({ timeoutMs })` (default ~10.6s, derived from post-handshake wait + handshake delay + TCP connect)
- **Overall retry deadline** (`CONNECTION_OVERALL_TIMEOUT_MS`) aligned with connect-phase limits so one successful connect is not cut short by a shorter overall cap
- Removed fixed `CONNECTION_MAX_RETRIES` (15)

**Retry loop (`tunnel/index.ts`)**
- Deadline-based retries using `performance.now()` (monotonic, not affected by clock skew)
- Fresh `RemoteXpcConnection` per attempt (avoids stale socket/listener state after failed `connect()`)
- **500ms backoff** after fast failures (attempt &lt; 500ms), capped by remaining overall time — reduces load on an unavailable RSD endpoint without delaying slow failures
- Logs remaining budget (`…ms left`) instead of retry count

**Refactor (`remote-xpc-connection.ts`)**
- Split `connect()` / `close()` into focused helpers (`ConnectSession`, socket handlers, handshake, graceful shutdown)
- `noDelay` / `keepAlive` passed via `net.connect()` options

| Before | After |
|--------|-------|
| Up to ~52s on repeated failures | ~10.6s overall cap |
| 3s per failed attempt | 500ms if TCP never connects |
| 15 fixed retries + 500ms delay always | Retries until deadline; backoff only on fast failures |

### DIFF
--- a/src/lib/remote-xpc/remote-xpc-connection.ts
+++ b/src/lib/remote-xpc/remote-xpc-connection.ts
@@ -243,7 +243,7 @@ class RemoteXpcConnection {
     let accumulatedData = Buffer.alloc(0);
     let partialExtractionTimer: NodeJS.Timeout | undefined;
 
-    const schedulePartialExtractionTimeout = (data: Buffer): void => {
+    const schedulePartialExtractionTimeout = (): void => {
       if (partialExtractionTimer) {
         return;
       }
@@ -253,7 +253,7 @@ class RemoteXpcConnection {
         log.warn(
           'Service extraction timeout reached, resolving with current data',
         );
-        const finalResponse = extractServices(data.toString('utf8'));
+        const finalResponse = extractServices(accumulatedData.toString('utf8'));
         session.settleSuccess(finalResponse);
       }, SERVICE_EXTRACTION_TIMEOUT_MS);
       session.trackPartialExtractionTimer(partialExtractionTimer);
@@ -299,7 +299,7 @@ class RemoteXpcConnection {
   private tryResolveServicesFromPayload(
     session: ConnectSession,
     accumulatedData: Buffer,
-    schedulePartialExtractionTimeout: (data: Buffer) => void,
+    schedulePartialExtractionTimeout: () => void,
   ): void {
     const dataStr = accumulatedData.toString('utf8');
     if (!dataStr.includes('com.apple') || !dataStr.includes('Port')) {
@@ -313,7 +313,7 @@ class RemoteXpcConnection {
         return;
       }
 
-      schedulePartialExtractionTimeout(accumulatedData);
+      schedulePartialExtractionTimeout();
     } catch (error) {
       log.warn(
         `Error extracting services: ${error}, continuing to collect data`,
--- a/src/lib/remote-xpc/remote-xpc-connection.ts
+++ b/src/lib/remote-xpc/remote-xpc-connection.ts
@@ -86,9 +86,7 @@ class RemoteXpcConnection {
    * Connect to the remote device and perform handshake
    * @returns Promise that resolves with the list of available services
    */
-  async connect(
-    options?: RemoteXpcConnectOptions,
-  ): Promise<ServicesResponse> {
+  async connect(options?: RemoteXpcConnectOptions): Promise<ServicesResponse> {
     if (this._isConnected) {
       throw new Error('Already connected');
     }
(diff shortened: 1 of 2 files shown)


src/lib/remote-xpc/remote-xpc-connection.ts
```diff
@@ -6,15 +6,34 @@ import Handshake from './handshake.js';
...
+ * Must cover the longest connect-phase timeout plus earlier phases.
+ */
+export const CONNECTION_DEFAULT_OPERATION_TIMEOUT_MS = Math.max(
+  SERVICE_EXTRACTION_TIMEOUT_MS,
+  SERVICE_AFTER_HANDSHAKE_TIMEOUT_MS +
+    HANDSHAKE_DELAY_MS +
+    CONNECTION_CONNECT_TIMEOUT_MS,
+);
+/** TunnelManager retry budget; never shorter than a single connect attempt. */
+export const CONNECTION_OVERALL_TIMEOUT_MS =
```
harsha509, line 27: Just conforming the intention here:
so calculating all. `CONNECTION_OVERALL_TIMEOUT_MS = evals to 10600ms`

if a conn stalls for fails after ~ 10 seconds , `performance.now() > deadline` evaluates to true  = means **slow failure results in exactly 0 retries.**

src/lib/tunnel/index.ts
```diff
@@ -86,18 +92,12 @@ class TunnelManagerService {
         } catch (error) {
           lastError = error;
           log.warn(
-            `RemoteXPC connection attempt failed (${retries} retries left): ${error}`,
+            `RemoteXPC connection attempt failed (${remainingMs}ms left): ${error}`,
```
harsha509: nit:
i see `remainingMs` is captured before `await remoteXPC.connect()` runs. 
If the connection fails after several seconds, the log warns with the stale pre-attempt budget

src/lib/remote-xpc/remote-xpc-connection.ts
```diff
@@ -325,6 +163,274 @@ class RemoteXpcConnection {
...
+      if (partialExtractionTimer) {
+        return;
+      }
+
+      partialExtractionTimer = setTimeout(() => {
+        partialExtractionTimer = undefined;
+        log.warn(
+          'Service extraction timeout reached, resolving with current data',
+        );
+        const finalResponse = extractServices(data.toString('utf8'));
```
harsha509: may be a closure bug here because `Buffer.concat()` creates a new `buffer` on every chunk. 
capturing `data` as a parameter here means this 5 second timer holds a stale snapshot. A
any chunks that arrive after the timer starts will be ignored when the timer fires.

src/lib/remote-xpc/remote-xpc-connection.ts
mykola-mokhnach: yes, this is the worst case if every connection stage times out and fails
would you prefer to reduce stage timeouts to fit more retries?

src/lib/remote-xpc/remote-xpc-connection.ts
mykola-mokhnach: Good catch. The timer was freezing accumulatedData at schedule time via the data parameter.

It now closes over the accumulatedData binding in registerConnectSocketHandlers, so when the 5s timer fires it parses whatever has been concatenated by then—including chunks that arrived after the timer was started. Incoming data events still run tryResolveServicesFromPayload on each chunk, so if services become parseable before the timer fires, settleSuccess wins as before.

src/lib/tunnel/index.ts
mykola-mokhnach: changed it to use after-attempt timestamp

src/lib/remote-xpc/remote-xpc-connection.ts
harsha509: I think its oky mykola, we can keep as it is
if it hands for ~ 10 seconds it is dead anyway. we are good as long as fast failyres retires

harsha509 (approved):

PR #218, opened 2026-05-21
fix: Partial XPC services discovery

### LINKED ISSUE #215: RSD service discovery truncates the service catalog (intermittent "Service ... not found")
## Summary

`RemoteXpcConnection` service discovery can settle on a **partial** RSD service catalog, so `findService()` throws `Service <name> not found` for services that the device actually exposes. Observed with `com.apple.mobile.diagnostics_relay.shim.remote` (diagnostics integration test `before` hook), while other service tests pass on the same device/tunnel.

## Root cause

In `src/lib/remote-xpc/remote-xpc-connection.ts`:

1. After the handshake, the device sends its whole service catalog as one large XPC message (length-prefixed XPC wrapper, magic `0x29b00b92` — a parser already exists in `xpc-protocol.ts`).
2. `registerConnectSocketHandlers` instead listens to raw `socket.on('data')` and, on **every** chunk, calls `tryResolveServicesFromPayload` → `extractServices`, a regex scrape (`/com\.apple(?:\.[\w-]+)+/g`) over the raw, possibly-incomplete UTF-8 bytes.
3. The instant that scrape yields `services.length > 0`, `session.settleSuccess` freezes `this._services` and ignores all further data.

The catalog spans multiple TCP segments, so `_services` is frozen to whatever was in the first segment(s). Services later in the byte stream never make the list.

## Why it looks intermittent

The cutoff is MTU / timing / catalog-order dependent. It flips pass↔fail with no code change — an iOS update adding services, a tunnel MSS change, or different TCP-segment coalescing. Services like `afc` / `installation_proxy` / `springboard` pass because they happen to sit earlier in the catalog.

## History

- Settle-on-first-non-empty logic introduced in #24 (`b918bf1`).
- #214 (`af8b81a`) rewrote the path into `tryResolveServicesFromPayload` but carried the flawed rule forward unchanged.
- Not introduced by any recent branch — pre-existing on `main`.

## Confirm

Log the discovered list before `findService`:
```ts
const svcs = remoteXPC.getServices();
console.log(svcs.length, svcs.map(s => s.serviceName));
```
A short list (~10–20) missing diagnostics on a device that exposes 40+ confirms truncation.

## Fix options

- **Proper:** reassemble the HTTP/2 DATA frame(s), parse the XPC wrapper length, run `extractServices` once on the complete message.
- **Mitigation:** stop settling on first non-empty extract — debounce until a quiet gap (~200ms no new `data`) or the post-handshake timeout, then extract once on the full buffer.

## Acceptance criteria

- [ ] Service discovery returns the complete catalog deterministically
- [ ] `findService` for any device-exposed service no longer fails intermittently
- [ ] Regression test covering a multi-segment catalog payload

mykola-mokhnach:
## Summary

Fixes [#215](https://github.com/appium/appium-ios-remotexpc/issues/215): RSD service discovery could settle on a **partial** service catalog, causing intermittent `Service <name> not found` errors (e.g. `com.apple.mobile.diagnostics_relay.shim.remote`) even when the device exposes the service.

**Root cause:** After the HTTP/2 handshake, `RemoteXpcConnection` scraped raw socket bytes with a regex and resolved as soon as any `com.apple.*` names appeared. The catalog arrives as one large XPC message split across multiple TCP segments, so the first chunk often contained only early services (`afc`, `installation_proxy`, `springboard`, …) while later entries were never included.

**Approach (aligned with [pymobiledevice3 `RemoteXPCConnection`](https://github.com/doronz88/pymobiledevice3/blob/master/pymobiledevice3/remote/remotexpc.py)):**

- Parse incoming **HTTP/2 DATA frames** incrementally (`http2-frame-parser.ts`) instead of treating the socket as UTF-8 text.
- **Reassemble** partial XPC payloads before decode (`ServiceCatalogCollector`), matching pymobiledevice3’s `_previous_frame_data` pattern.
- Extract services from the decoded `Services` dictionary (`servicesFromXpcBody`) instead of regex matching.
- Send **WINDOW_UPDATE** on even streams when receiving data so the device can finish sending the catalog.
- Fix **XPC dictionary/array decode** to read length-prefixed payloads (required for real handshake messages).

Removes the “settle on first non-empty scrape” behavior and the partial-extraction timeout fallback.

## Test plan

- [x] `npm run test:unit -- --grep "RSD service catalog"` — partial XPC, multi-chunk catalog, split HTTP/2 frame
- [x] `npm run test:unit` — full unit suite (410 tests)
- [x] `npm run build` && `npm run lint`
- [ ] Manual: connect to a device/tunnel and log `remoteXPC.getServices().length` — expect full catalog (40+ services), including `com.apple.mobile.diagnostics_relay.shim.remote`
- [ ] Manual: `npm run test:diagnostics` (or other previously flaky RSD service tests) on the same device/tunnel

### DIFF
--- a/src/lib/remote-xpc/http2-frame-parser.ts
+++ b/src/lib/remote-xpc/http2-frame-parser.ts
@@ -1,4 +1,4 @@
-import { WindowUpdateFrame } from './handshake-frames.js';
+import { InvalidDataError, WindowUpdateFrame } from './handshake-frames.js';
 
 const FRAME_HEADER_SIZE = 9;
 const FRAME_TYPE_DATA = 0x00;
@@ -68,14 +68,35 @@ function parseFrame(buffer: Buffer): ParsedFrame {
     return { type: 'other' };
   }
 
-  let data = body;
-  if (flags & FLAG_PADDED) {
-    const padLength = body[0] ?? 0;
-    data = body.subarray(1, body.length - padLength);
-  }
+  const data = stripDataFramePadding(body, flags);
 
   return {
     type: 'data',
     frame: { streamId, data, bodyLen: length },
   };
+}
+
+/**
+ * Strip HTTP/2 DATA frame padding (RFC 7540 §6.1).
+ * @throws {InvalidDataError} when padding length is invalid
+ */
+function stripDataFramePadding(body: Buffer, flags: number): Buffer {
+  if (!(flags & FLAG_PADDED)) {
+    return body;
+  }
+
+  if (body.length === 0) {
+    throw new InvalidDataError(
+      'PROTOCOL_ERROR: PADDED DATA frame has empty payload',
+    );
+  }
+
+  const padLength = body.readUInt8(0);
+  if (padLength >= body.length) {
+    throw new InvalidDataError(
+      'PROTOCOL_ERROR: Padding exceeds frame size',
+    );
+  }
+
+  return body.subarray(1, body.length - padLength);
 }
--- a/src/lib/remote-xpc/http2-frame-parser.ts
+++ b/src/lib/remote-xpc/http2-frame-parser.ts
@@ -1,0 +1,84 @@
+import { WindowUpdateFrame } from './handshake-frames.js';
+
+const FRAME_HEADER_SIZE = 9;
+const FRAME_TYPE_DATA = 0x00;
+const FLAG_PADDED = 0x08;
+
+export interface ParsedDataFrame {
+  readonly streamId: number;
+  readonly data: Buffer;
+  readonly bodyLen: number;
+}
+
+export type ParsedFrame =
+  | { readonly type: 'data'; readonly frame: ParsedDataFrame }
+  | { readonly type: 'other' };
+
+/**
+ * Incrementally parse HTTP/2 frames from a byte stream (RFC 7540).
+ */
+export class Http2FrameParser {
+  private buffer: Buffer = Buffer.alloc(0);
+
+  append(chunk: Buffer): ParsedFrame[] {
+    this.buffer = Buffer.concat([this.buffer, chunk]);
+    const frames: ParsedFrame[] = [];
+
+    while (this.buffer.length >= FRAME_HEADER_SIZE) {
+      const length =
+        (this.buffer[0] << 16) |
+        (this.buffer[1] << 8) |
+        this.buffer[2];
+      const totalSize = FRAME_HEADER_SIZE + length;
+      if (this.buffer.length < totalSize) {
+        break;
+      }
+
+      const frameBuffer = this.buffer.subarray(0, totalSize);
+      this.buffer = this.buffer.subarray(totalSize);
+      frames.push(parseFrame(frameBuffer));
+    }
+
+    return frames;
+  }
+}
+
+/**
+ * Emit WINDOW_UPDATE frames for even-numbered streams, matching `remoted` behavior.
+ */
+export function buildWindowUpdateFrames(
+  streamId: number,
+  increment: number,
+): Buffer[] {
+  if (streamId % 2 !== 0 || increment <= 0) {
+    return [];
+  }
+  return [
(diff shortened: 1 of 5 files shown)


src/lib/remote-xpc/service-catalog.ts
```diff
@@ -0,0 +1,76 @@
...
+  /**
+   * Feed one DATA frame payload. Returns the catalog when a complete XPC
+   * message containing `Services` has been decoded.
+   */
+  ingestDataPayload(chunk: Buffer): ServicesResponse | null {
+    const combined = Buffer.concat([this.previousFrameData, chunk]);
+
+    try {
+      const message = decodeMessage(combined);
+      this.previousFrameData = Buffer.alloc(0);
```
harsha509: silently truncatng trailing bytes ?

src/lib/remote-xpc/http2-frame-parser.ts
```diff
@@ -0,0 +1,81 @@
...
+  const body = buffer.subarray(FRAME_HEADER_SIZE, FRAME_HEADER_SIZE + length);
+
+  if (type !== FRAME_TYPE_DATA) {
+    return { type: 'other' };
+  }
+
+  let data = body;
+  if (flags & FLAG_PADDED) {
+    const padLength = body[0] ?? 0;
+    data = body.subarray(1, body.length - padLength);
```
harsha509: what happens on server glicth? should we do 

if (padLength >= body.length) {
  throw new Error('PROTOCOL_ERROR: Padding exceeds frame size');
}

src/lib/remote-xpc/remote-xpc-connection.ts
```diff
@@ -11,14 +20,11 @@ export const CONNECTION_CONNECT_TIMEOUT_MS = 3_000;
...
 /** Wait for service list after handshake before failing the connect attempt. */
 const SERVICE_AFTER_HANDSHAKE_TIMEOUT_MS = 10_000;
-/** Resolve with partial data if service parsing stalls mid-stream. */
-const SERVICE_EXTRACTION_TIMEOUT_MS = 5_000;
 /**
  * Default max time for one connect attempt (TCP + handshake + service discovery).
  * Must cover the longest connect-phase timeout plus earlier phases.
  */
 export const CONNECTION_DEFAULT_OPERATION_TIMEOUT_MS = Math.max(
-  SERVICE_EXTRACTION_TIMEOUT_MS,
```
harsha509, line 21: so as `SERVICE_EXTRACTION_TIMEOUT_MS` removed, math.max is redundant here.

harsha509 (approved):

src/lib/remote-xpc/service-catalog.ts
mykola-mokhnach: Yes — that was a real bug.

What was wrong

decodeMessage(combined) only reads one XPC wrapper, but it never reported how many bytes were used. Clearing previousFrameData after a successful decode dropped any extra bytes in the same chunk (e.g. a second XPC message or the start of the catalog).

There was a related issue in getXpcMessageByteLength: bodyLen was read at offset 12 instead of 8, so length checks were wrong and a truncated buffer could look like a complete 24-byte empty message.

src/lib/remote-xpc/http2-frame-parser.ts
mykola-mokhnach: remoted on iOS rarely sends padded DATA for the catalog, but validating is cheap and matches what a real HTTP/2 stack would do on PROTOCOL_ERROR

stripDataFramePadding() now throws InvalidDataError when:

- PADDED is set but the payload is empty
- padLength >= body.length

processIncomingData catches parser errors and calls session.settleFailure, so a server glitch fails the connect attempt instead of an unhandled data exception.

src/lib/remote-xpc/remote-xpc-connection.ts
mykola-mokhnach: oops, removed

PR #219, opened 2026-05-21
fix: AFC connection desync

### LINKED ISSUE #216: AFC: a timed-out / abandoned read permanently desyncs the connection (`Invalid AFC magic`)
## Summary

The AFC codec cannot recover from an abandoned read. If a `readExact` call times out — or an in-flight operation is abandoned for any other reason — the bytes of that operation's (late-arriving) response are still appended to the shared socket buffer but never consumed. Every subsequent operation then reads stale, misaligned bytes and fails with `Invalid AFC magic`. The connection is corrupted permanently.

## Root cause

`src/services/ios/afc/codec.ts`

AFC is a raw byte stream with no request/response correlation — `readExact(n)` simply returns the next `n` bytes off the socket.

On timeout, `readExact` only removes its own waiter and rejects. It does **not** discard the pending bytes and does **not** destroy the socket:

```js
waiter.timer = setTimeout(() => {
  const idx = state.waiters.indexOf(waiter);
  if (idx >= 0) {
    state.waiters.splice(idx, 1);
    reject(new Error(`readExact timeout after ${timeoutMs}ms`));
  }
}, timeoutMs);
```

The `onData` handler keeps appending the late response to `state.buffer`. The next `readExact` sees `buffer.length >= n` and returns those stale bytes as if they were a fresh 40-byte header → `Invalid AFC magic`. The "magic" bytes decode to AFC payload fragments such as `st_size` / `st_ifmt`, confirming the read is mid-payload of a previous response.

## How it surfaces

`AFC_OPERATION_TIMEOUT_MS` (15s) is the default read timeout. Any operation slower than that — or any caller that abandons an operation mid-flight — poisons a shared `AfcService` connection for every later operation on it.

Observed in the `test/integration/afc-test.ts` suite: one slow test left the single shared `AfcService` connection desynced, and the next three tests all failed with `Invalid AFC magic`. The test trigger is being addressed separately (scoping `walk` to a subdirectory), but that only avoids tripping this bug — it does not fix it.

## Suggested fix

Make an abandoned/timed-out read fatal for the connection instead of silently corrupting it:

- On `readExact` timeout, `destroy()` the socket (and/or mark the `AfcService` instance dead) so it cannot be silently reused.
- Reject any subsequent operation on a dead connection with a clear error.

Draining `state.buffer` alone is not sufficient — the late response may still be partially in flight — so tearing the connection down is the safe choice.

mykola-mokhnach:
## Summary

Fixes [#216](https://github.com/appium/appium-ios-remotexpc/issues/216).

AFC uses a sequential byte stream with no request/response correlation. When `readExact` timed out, it removed the waiter but kept appending late response bytes to the shared buffer. The next operation then read misaligned data and failed with `Invalid AFC magic`, permanently poisoning the connection.

This change makes abandoned or timed-out reads fatal for the connection:

- **`readExact` timeout** — tears down buffered-reader state, destroys the socket, and rejects with `AfcConnectionError`.
- **Invalid AFC magic** — treated as stream desync; same fatal teardown.
- **`AfcService`** — records a terminal error on `AfcConnectionError` so later operations on the same instance fail immediately instead of silently reusing a desynced socket.
- **`AfcConnectionError`** — exported for callers that need to detect a dead connection and create a new `AfcService`.

## Test plan

- [x] `NODE_ENV=test npx mocha 'test/unit/afc/**/*.ts'`
  - New: timeout destroys socket; late bytes cannot poison the next read
  - Existing codec utility tests still pass
- [ ] `npm run lint`
- [ ] `npm run test:afc` (integration, requires device/tunnel)

### DIFF
--- a/src/services/ios/afc/codec.ts
+++ b/src/services/ios/afc/codec.ts
@@ -8,6 +8,7 @@ import {
   AFC_OPERATION_TIMEOUT_MS,
   NULL_BYTE,
 } from './constants.js';
+import { AfcConnectionError } from './errors.js';
 import { AfcError, AfcFopenMode, AfcOpcode } from './enums.js';
 
 export interface AfcHeader {
@@ -94,15 +95,30 @@ export function encodeHeader(
 }
 
 const SOCKET_STATES = new WeakMap<net.Socket, SocketState>();
+const FATAL_SOCKETS = new WeakSet<net.Socket>();
+
+/**
+ * Tear down buffered-reader state and destroy the socket after a fatal read error.
+ * Late-arriving bytes must not be left for a subsequent readExact on the same socket.
+ */
+export function fatalizeAfcSocket(socket: net.Socket, error: Error): void {
+  FATAL_SOCKETS.add(socket);
+  cleanupSocketState(socket, error);
+  if (!socket.destroyed) {
+    socket.destroy();
+  }
+}
 
 /**
  * Read exactly `n` bytes from a socket, waiting up to timeoutMs.
+ * On timeout the socket is destroyed; further reads throw AfcConnectionError.
  */
 export async function readExact(
   socket: net.Socket,
   n: number,
   timeoutMs = 30000,
 ): Promise<Buffer> {
+  assertSocketReadable(socket);
   const state = ensureSocketState(socket);
 
   if (state.buffer.length >= n) {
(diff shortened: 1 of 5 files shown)


src/services/ios/afc/codec.ts
```diff
@@ -94,15 +95,30 @@ export function encodeHeader(
...
  * Read exactly `n` bytes from a socket, waiting up to timeoutMs.
+ * On timeout the socket is destroyed; further reads throw AfcConnectionError.
  */
 export async function readExact(
   socket: net.Socket,
   n: number,
   timeoutMs = 30000,
 ): Promise<Buffer> {
+  assertSocketReadable(socket);
   const state = ensureSocketState(socket);
```
harsha509, line 121: readExtract is calling ` assertSocketReadable(socket);` and ensureSocketState(socket) calling `assertSocketReadable(socket);`

we can keep one instead removing redundant one

harsha509 (approved):

src/services/ios/afc/codec.ts
mykola-mokhnach: only keeping ensureSocketState

PR #220, opened 2026-05-21
fix: Respect local file system limitations while pulling remote files

### LINKED ISSUE #115: afc: sanitize file names while pulling to local (handle OS dependent bad chars)
> should we also think about cases where the local file system does not like some chars of the remote one? E.g. if on windows we try to pull a remote file that contains `>`? 

 _Originally posted by @mykola-mokhnach in [#113](https://github.com/appium/appium-ios-remotexpc/pull/113/changes/BASE..b89eed117fe29943b389b7c52dfab44ae552588d#r2627140727)_

mykola-mokhnach:
## Summary

Fixes [#115](https://github.com/appium/appium-ios-remotexpc/issues/115) by sanitizing remote AFC path segments for the **host** filesystem during `pull`, and disambiguating names only when a real clash would occur in the destination directory.

- **`sanitize-local-filename.ts`** — Platform-specific sanitization (ported from `sanitize-filename` logic, no extra dependency):
  - **Windows**: illegal chars, control chars, reserved names (`CON`, …), trailing dots/spaces, 255-byte UTF-8 limit
  - **macOS**: `/`, `:`, control chars, `.` / `..`
  - **Linux**: `/`, control chars, `.` / `..` (keeps chars like `>` that are valid locally)
- **`pull-local-name-allocator.ts`** — Per-`pull()` allocator that:
  - Uses sanitized names by default
  - Adds `_<8-char-uuid>` **only on clash** (another file in the same pull, or an existing local path)
  - Respects volume case rules via `isCaseSensitiveDirectory`
- **`local-filesystem-case.ts`** — Case-sensitivity detection for clash comparison:
  - **macOS**: `df` → `diskutil info -plist` → `plist.parsePlist` (`FilesystemName`, `FilesystemUserVisibleName`, …)
  - Falls back to case-insensitive with a log message if detection fails
  - **Windows**: case-insensitive; **Linux**: case-sensitive
- **`AfcService.pull` / `_pullRecursiveInternal`** — Wire allocator for single-file and recursive pulls; remote paths and `match` globs still use original device names

Bumps `@appium/support` to `^7.2.2` for `plist.parsePlist`.

## Test plan

- [x] Unit: `sanitize-local-filename` (win32 / darwin / linux rules)
- [x] Unit: `pull-local-name-allocator` (no suffix without clash; suffix on sanitize collision and on-disk collision; case-sensitive vs case-insensitive volumes)
- [x] Unit: `local-filesystem-case` (plist parser fixtures; live `diskutil info -plist` on macOS)
- [ ] Integration (device): pull a file with host-invalid chars (e.g. `report>file.txt` on Windows) and confirm it lands locally
- [ ] Integration (device): pull two remote files that sanitize to the same name (e.g. `file?` and `*file*`) and confirm both are written with distinct local names
- [ ] Integration (optional): recursive directory pull with mixed safe/unsafe entry names

### DIFF
--- a/src/services/ios/afc/index.ts
+++ b/src/services/ios/afc/index.ts
@@ -397,10 +397,15 @@ export class AfcService {
     const pullSingleFile = async (
       remoteFilePath: string,
       localFilePath: string,
+      nameAllocated = false,
     ): Promise<void> => {
       log.debug(`Pulling file from '${remoteFilePath}' to '${localFilePath}'`);
 
-      if (!overwrite && (await this._localPathExists(localFilePath))) {
+      if (
+        !nameAllocated &&
+        !overwrite &&
+        (await this._localPathExists(localFilePath))
+      ) {
         throw new Error(`Local file already exists: ${localFilePath}`);
       }
 
@@ -428,7 +433,7 @@ export class AfcService {
           )
         : localDst;
 
-      await pullSingleFile(remoteSrc, targetPath);
+      await pullSingleFile(remoteSrc, targetPath, localDstIsDirectory);
       return;
     }
 
@@ -716,10 +721,6 @@ export class AfcService {
         await ensureLocalDir();
 
         const targetPath = path.join(localDirPath, localEntryName);
-        if (!overwrite && (await this._localPathExists(targetPath))) {
-          throw new Error(`Local file already exists: ${targetPath}`);
-        }
-
         await this._pullFile(entryPath, targetPath);
 
         if (onPullProgress) {
--- a/src/services/ios/afc/index.ts
+++ b/src/services/ios/afc/index.ts
@@ -32,6 +32,7 @@ import {
 } from './constants.js';
 import { AfcError, AfcFileMode, AfcOpcode } from './enums.js';
 import { AfcConnectionError } from './errors.js';
+import { PullLocalNameAllocator } from './pull-local-name-allocator.js';
 import { createAfcReadStream, createAfcWriteStream } from './stream-utils.js';
 
 const log = getLogger('AfcService');
@@ -388,6 +389,10 @@ export class AfcService {
       throw new Error(`Remote path does not exist: ${remoteSrc}`);
     }
 
+    const localNames = new PullLocalNameAllocator((localPath) =>
+      this._localPathExists(localPath),
+    );
+
     const pullSingleFile = async (
       remoteFilePath: string,
       localFilePath: string,
@@ -408,15 +413,18 @@ export class AfcService {
     const isDir = await this.isdir(remoteSrc);
 
     if (!isDir) {
-      const baseName = path.posix.basename(remoteSrc);
+      const remoteBaseName = path.posix.basename(remoteSrc);
 
-      if (match && !minimatch(baseName, match)) {
+      if (match && !minimatch(remoteBaseName, match)) {
         return;
       }
 
       const localDstIsDirectory = await this._isLocalDirectory(localDst);
       const targetPath = localDstIsDirectory
-        ? path.join(localDst, baseName)
+        ? path.join(
+            localDst,
+            await localNames.allocate(localDst, remoteBaseName),
+          )
         : localDst;
 
       await pullSingleFile(remoteSrc, targetPath);
@@ -435,6 +443,7 @@ export class AfcService {
       match,
       overwrite,
       callback: onPullProgress,
+      localNames,
     });
   }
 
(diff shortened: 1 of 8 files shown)


src/services/ios/afc/pull-local-name-allocator.ts
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    throw new Error(
+      `Could not allocate a unique local name for remote segment '${remoteSegment}'`,
+    );
+  }
+
+  private async hasClash(parentDir: string, name: string): Promise<boolean> {
+    if (await this.isUsedInPull(parentDir, name)) {
+      return true;
```
harsha509, line 51: this may override `overwrite`

src/services/ios/afc/index.ts
```diff
@@ -690,7 +714,7 @@ export class AfcService {
 
         await ensureLocalDir();
 
-        const targetPath = path.join(localDirPath, entry);
+        const targetPath = path.join(localDirPath, localEntryName);
         if (!overwrite && (await this._localPathExists(targetPath))) {
```
harsha509: is this dead code now ? as `targetPath` is always a freshly allocatd

src/services/ios/afc/sanitize-local-filename.ts
```diff
@@ -0,0 +1,144 @@
...
+const POSIX_PATH_SEP_RE = /\//g;
+
+/** Colon is invalid in macOS / HFS+ / APFS file names. */
+const DARWIN_ILLEGAL_RE = /:/g;
+
+/** Unix reserved single-segment names (`.` and `..`). */
+const RESERVED_DOTS_RE = /^\.+$/;
+
+/** Windows reserved device names (case-insensitive, optional extension). */
+const WINDOWS_RESERVED_RE = /^(con|prn|aux|nul|com[0-9]|lpt[0-9])(\..*)?$/i;
```
harsha509, line 25: i assume win reserves `COM1–COM9 and LPT1–LPT9 ` only `COM0/LPT0 are valid filenames`
may be best to use `com[1-9]/lpt[1-9]` ?

src/services/ios/afc/sanitize-local-filename.ts
```diff
@@ -0,0 +1,144 @@
...
+  const dot = filename.lastIndexOf('.');
+  if (dot > 0) {
+    return `${filename.slice(0, dot)}_${suffix}${filename.slice(dot)}`;
+  }
+  return `${filename}_${suffix}`;
+}
+
+/** Sanitized name with a disambiguating suffix, still within host filename byte limits. */
+export function appendUniqueSuffix(sanitized: string, suffix: string): string {
+  const suffixed = withUniqueSuffix(sanitized, suffix);
```
harsha509: On 255-byte names truncation can eat the UUID suffix, should be trim before appendng ?

src/services/ios/afc/pull-local-name-allocator.ts
mykola-mokhnach: yes, now it does not

Changed PullLocalNameAllocator to take overwrite from pull():

- isUsedInPull — still always a clash (two entries in one pull must not target the same local name).
- pathExists — only a clash when overwrite is false; when true, an existing file is left to the normal pull overwrite behavior.

src/services/ios/afc/index.ts
mykola-mokhnach: Yes — for the recursive path it was redundant.

With overwrite: false, allocate() already loops until pathExists(parentDir/candidate) is false (and in-pull names do not collide). So targetPath should not exist on disk at allocation time, and the check at 718–719 could not fail in normal operation (only a TOCTOU race).

src/services/ios/afc/sanitize-local-filename.ts
mykola-mokhnach: there are caveats:

Some APIs or legacy software may still treat them specially.

Behavior can differ between:
- Win32 APIs
- NT native paths (\\?\)
- Explorer vs command shell
- network shares
- Certain tools may reject them defensively even though NTFS itself allows them.

sanitize-filename does exclude them, so we do as well

src/services/ios/afc/sanitize-local-filename.ts
mykola-mokhnach: makes sense. fixed

src/services/ios/afc/sanitize-local-filename.ts
mykola-mokhnach: better safe than sorry :)

harsha509 (approved):

harsha509 (approved):
Thank you Mykola...!

PR #221, opened 2026-05-22
feat: Unify dnssd discovery

### LINKED ISSUE #206: Replace dns-sd CLI / dnssd library backends with a raw mDNS browser for cross-platform Apple TV discovery
## Context
The current discovery implementation has two backends:

BonjourDiscoveryBackend — wraps the macOS dns-sd CLI. Required because Apple's mDNSResponder advertises non-RFC-6335 service names (e.g. _remotepairing-manual-pairing._tcp, 28 chars) that strict pure-JS mDNS libraries reject.
DnssdDiscoveryBackend — uses the dnssd npm library on Linux/Windows. However, dnssd enforces RFC 6335 (max 15-char service names), so it cannot see Apple-style long names anywhere either.
Net effect: discovery only works on macOS today.

## Reference implementation
pymobiledevice3 solves this with a dependency-light raw mDNS browser ([pymobiledevice3/bonjour.py](https://github.com/doronz88/pymobiledevice3/blob/master/pymobiledevice3/bonjour.py)):

- Opens raw UDP sockets on port 5353 for IPv4 + IPv6 multicast.
- Sends raw DNS-SD PTR queries.
- Parses incoming mDNS packets directly (PTR / SRV / TXT / A / AAAA).

No third-party mDNS library, no service-name length validation — works on macOS / Linux / Windows for any service name.
Proposal

- One backend, one code path on all platforms.
- No dependency on the macOS dns-sd CLI or the dnssd npm library.
- Apple's long service names work everywhere.

mykola-mokhnach:
## Summary

Closes [#206](https://github.com/appium/appium-ios-remotexpc/issues/206).

Replaces the split discovery stack (macOS `dns-sd` CLI + cross-platform `dnssd` npm package) with a single **raw mDNS browser** on UDP port 5353. Apple’s non–RFC-6335 service names (e.g. `_remotepairing-manual-pairing._tcp`) now work on **macOS, Linux, and Windows** without platform-specific backends.

**Removed**
- `BonjourDiscoveryBackend` (`dns-sd`)
- `DnssdDiscoveryBackend` (`dnssd` / `@types/dnssd`)
- `bonjour-output-parsers.ts` and its unit tests

**Added**
- `mdns-protocol.ts` — DNS encode/decode, PTR query building, response parsing
- `mdns-browser.ts` — multicast sockets, browse loop
- `MdnsDiscoveryBackend` — maps browse results to `DiscoveredDevice`
- Fixture-based e2e test with `MdnsTestResponder`
- Unit tests for protocol parsing and factory wiring

`createDiscoveryBackend()` always returns `MdnsDiscoveryBackend`. Pairing and tunnel services are unchanged at the API level; only the transport underneath discovery changed.

## Motivation

- `dnssd` enforces RFC 6335 (15-char service names) and cannot see Apple’s long DNS-SD types.
- `dns-sd` works on macOS only.
- Net effect before this change: reliable discovery only on macOS for manual pairing.

## Test plan

- [x] `npm run build`
- [x] `npm run lint`
- [x] `npm run test:unit` (432 tests)
- [x] `npm run test:discovery` — fixture e2e (local mDNS responder on port 5353)
- [ ] Optional live LAN test:
  ```bash
  REMOTE_PAIRING_LIVE_DISCOVERY=1 npm run test:discovery
  ```
  Discovers any Apple device advertising `_remotepairing._tcp` (Mac, Apple TV, etc.), not Apple TV only.

## Notes

- Binding to UDP **5353** may require elevated privileges on some hosts (`sudo`), same as the old `dnssd` path.
- Live discovery is opt-in via `REMOTE_PAIRING_LIVE_DISCOVERY=1` and skipped by default in CI.


mykola-mokhnach:
@harsha509 could you please help with testing?

src/lib/discovery/mdns-browser.ts
```diff
@@ -0,0 +1,284 @@
...
+      hostAddresses.set(key, existing);
+    }
+  };
+
+  try {
+    const query = buildQuery(serviceTypeFqdn, QTYPE_PTR, false);
+    sendQueryAll(sockets, query);
+
+    await collectResponses(sockets, timeoutMs, (data) => {
+      for (const rr of parseMdnsMessage(data)) {
```
harsha509: `parseMdnsMessage` here can crash the process if there is any ma;formed packet

mykola-mokhnach: added try/catch

harsha509 (approved):
LGTM!

src/lib/discovery/mdns-browser.ts
```diff
@@ -0,0 +1,292 @@
...
+  }
+  return unique;
+}
+
+function sendQueryAll(handles: MdnsSocketHandle[], packet: Buffer): void {
+  for (const { socket, family } of handles) {
+    if (family !== 'ipv4') {
+      continue;
+    }
+    socket.send(packet, MDNS_PORT, MDNS_MCAST_V4);
```
navin772, line 271: I think we should pass an error callback here and log as warning

navin772 (approved):

src/lib/discovery/mdns-browser.ts
mykola-mokhnach: could you please elaborate on this, maybe provide a code example?

navin772:
```ts
socket.send(packet, MDNS_PORT, MDNS_MCAST_V4, (err) => {
  if (err) {
    log.warn(`Failed to send mDNS query: ${err.message}`);
  }
});
```

PR #224, opened 2026-05-25
feat: Fail fast if a tunnel registry is not running

mykola-mokhnach:
## Summary

When the tunnel registry was down or a UDID had no tunnel entry, `getTunnelForDevice` could still reach `connectRemoteXPCUnlocked` and wait ~13s on RSD. Lookup used a non-strict registry client (10s timeout) plus duplicate HTTP (`hasTunnel` then `getTunnelConnection`).

This PR adds a dedicated tunnel availability path that fails in under ~500ms and never opens RSD when the registry or tunnel entry is missing.

## What changed

- New `src/lib/tunnel/tunnel-availability.ts` centralizes registry checks:
  - Read registry port from strongbox
  - TCP probe `127.0.0.1:{port}` (300ms) before HTTP
  - Single strict `GET /remotexpc/tunnels/{udid}` (500ms) per device lookup
  - Same validated client path for `getAvailableDevices` (list registry)
- `TunnelAvailabilityError` (`ERR_TUNNEL_AVAILABILITY`) with optional `cause` for wrapped fetch/socket errors; probe messages include `host:port` (e.g. `Tunnel registry at 127.0.0.1:42314 is not reachable…`)
- Tunnel-specific constants moved to `src/lib/tunnel/constants.ts` (`TUNNEL_REGISTRY_HOST`, `TUNNEL_REGISTRY_API_BASE_PATH`, timeouts)
- `services.ts` simplified: `withRemoteXpcConnection` still calls `getTunnelForDevice`; `Services` re-exports `getTunnelForDevice` and `getAvailableDevices` for **appium-xcuitest-driver** (`remotexpc.Services.getTunnelForDevice`)
- Package root exports `TunnelAvailabilityError`, `getTunnelForDevice`, `getAvailableDevices` from the tunnel module

## Behavior

| Scenario | Before | After |
|----------|--------|-------|
| No strongbox port / registry not listening / UDID not in registry | Slow or ambiguous failure; could attempt RSD | `TunnelAvailabilityError` in ~300–500ms; **no** `connectRemoteXPCUnlocked` |
| Registry up, tunnel listed | Works | Unchanged |
| Stale registry entry, dead tunnel process | ~13s RSD timeout | Still possible on RSD (out of scope) |

## Tests

- `test/unit/tunnel/tunnel-availability.spec.ts` — missing port, TCP refuse, GET 404, GET 200
- Updated `tunnel-availability-error.spec.ts` and `start-services-cleanup.spec.ts` (lookup failure must not call `connectRemoteXPCUnlocked`)

## Follow-up

- **appium-xcuitest-driver**: treat `TunnelAvailabilityError` in syslog / optional paths (e.g. `ios-device-log.ts`) so RemoteXPC is skipped immediately when the registry is down

### DIFF
--- a/src/lib/tunnel/tunnel-availability.ts
+++ b/src/lib/tunnel/tunnel-availability.ts
@@ -1,0 +1,177 @@
+import { BaseItem, strongbox } from '@appium/strongbox';
+import * as net from 'node:net';
+
+import { TUNNEL_CONTAINER_NAME } from '../../constants.js';
+import type { TunnelRegistryEntry } from '../types.js';
+import {
+  TUNNEL_REGISTRY_API_BASE_PATH,
+  TUNNEL_REGISTRY_HOST,
+  TUNNEL_REGISTRY_HTTP_TIMEOUT_MS,
+  TUNNEL_REGISTRY_PORT_PROBE_TIMEOUT_MS,
+} from './constants.js';
+import {
+  TunnelApiClient,
+  type TunnelApiClientOptions,
+  type TunnelEndpoint,
+} from './tunnel-api-client.js';
+
+const TUNNEL_REGISTRY_PORT = 'tunnelRegistryPort';
+
+export class TunnelAvailabilityError extends Error {
+  readonly code = 'ERR_TUNNEL_AVAILABILITY';
+
+  constructor(message: string, options?: { cause?: unknown }) {
+    super(message, options);
+    this.name = new.target.name;
+    Object.setPrototypeOf(this, new.target.prototype);
+  }
+}
+
+/**
+ * Resolve tunnel registry metadata for a device (host, RSD port, packet stream).
+ * Strongbox port, TCP probe, then a single `GET /remotexpc/tunnels/{udid}`.
+ * @throws {TunnelAvailabilityError} When registry or tunnel for the UDID is unavailable.
+ */
+export async function getTunnelForDevice(
+  udid: string,
+): Promise<TunnelEndpoint> {
+  const client = await createValidatedStrictRegistryClient();
+
+  let entry: TunnelRegistryEntry | null;
+  try {
+    entry = await client.getTunnelByUdid(udid);
+  } catch (err) {
+    const detail = err instanceof Error ? err.message : String(err);
+    throw new TunnelAvailabilityError(
+      `Failed to reach tunnel registry: ${detail}`,
+      { cause: err },
+    );
+  }
+
+  if (!entry) {
+    throw new TunnelAvailabilityError(
+      `No tunnel found for device ${udid}. Please run the tunnel creation script first`,
+    );
+  }
+
+  return mapEntryToEndpoint(entry);
+}
+
+/**
+ * Returns the list of device UDIDs currently in the tunnel registry.
+ * Used to include tunnel-only devices (e.g. Apple TV over WiFi)
+ * in the "connected devices" list for session validation.
+ *
+ * @returns UDIDs when the registry is reachable (possibly empty).
+ * @throws {TunnelAvailabilityError} When the registry port is missing or unreachable.
+ */
+export async function getAvailableDevices(): Promise<string[]> {
+  const client = await createValidatedStrictRegistryClient();
+  try {
+    return await client.getAvailableDevices();
+  } catch (err) {
+    const detail = err instanceof Error ? err.message : String(err);
+    throw new TunnelAvailabilityError(
+      `Failed to reach tunnel registry: ${detail}`,
+      { cause: err },
+    );
+  }
+}
+
+/**
+ * Read the tunnel registry HTTP port from strongbox.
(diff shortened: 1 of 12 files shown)


src/lib/tunnel/tunnel-availability.ts
```diff
@@ -0,0 +1,177 @@
...
+): Promise<TunnelEndpoint> {
+  const client = await createValidatedStrictRegistryClient();
+
+  let entry: TunnelRegistryEntry | null;
+  try {
+    entry = await client.getTunnelByUdid(udid);
+  } catch (err) {
+    const detail = err instanceof Error ? err.message : String(err);
+    throw new TunnelAvailabilityError(
+      `Failed to reach tunnel registry: ${detail}`,
```
harsha509: this is getting wrapped twice i think 
becuase we do same in here inner client theowing https://github.com/appium/appium-ios-remotexpc/blob/a7b3c84d2399763fb078b90ba557c5e57aedda95/src/lib/tunnel/tunnel-api-client.ts#L104

src/lib/tunnel/tunnel-availability.ts
```diff
@@ -0,0 +1,177 @@
...
+  const tunnelRegistryPort = await item.read();
+  if (
+    tunnelRegistryPort === undefined ||
+    String(tunnelRegistryPort).trim() === ''
+  ) {
+    throw new TunnelAvailabilityError(
+      'Tunnel registry port not found. Please run the tunnel creation script first',
+    );
+  }
+  return Number.parseInt(String(tunnelRegistryPort), 10);
```
harsha509: can we validate port number like?
`if (!Number.isInteger(port) || port < 1 || port > 65535) throw new TunnelAvailabilityError('Tunnel registry port is invalid')`

src/lib/tunnel/tunnel-availability.ts
```diff
@@ -0,0 +1,177 @@
...
+}
+
+/**
+ * Verify the registry HTTP server is accepting TCP connections on localhost.
+ * @throws {TunnelAvailabilityError} When the port is closed or the probe times out.
+ */
+async function assertRegistryPortAcceptingConnections(
```
harsha509, line 86: as we are throwing `ECONNREFUSED`,  is tcp probing useful ?

src/lib/tunnel/tunnel-availability.ts
```diff
@@ -0,0 +1,177 @@
...
+  const unreachableMessage = `Tunnel registry at ${registryAddress} is not reachable. Please run the tunnel creation script first`;
+
+  await new Promise<void>((resolve, reject) => {
+    const socket = net.connect({ host: TUNNEL_REGISTRY_HOST, port });
+    const timer = setTimeout(() => {
+      socket.destroy();
+      reject(new TunnelAvailabilityError(unreachableMessage));
+    }, TUNNEL_REGISTRY_PORT_PROBE_TIMEOUT_MS);
+
+    const finish = (err?: Error) => {
```
harsha509, line 99: this is nice, are we calling it anywhere? i dont see it in `setTimeout` so

src/services.ts
```diff
@@ -40,21 +34,15 @@ import SyslogService from './services/ios/syslog-service/index.js';
...
-export class TunnelAvailabilityError extends Error {
-  readonly code = 'ERR_TUNNEL_AVAILABILITY';
-
-  constructor(message: string) {
-    super(message);
-    this.name = new.target.name;
-    Object.setPrototypeOf(this, new.target.prototype);
-  }
-}
+export {
```
harsha509, line 42: do we need export here too, i see module level exports alrady.

mykola-mokhnach: I keep it to ensure the backward compatibility. Removing the class from services would be a breaking change

src/lib/tunnel/tunnel-availability.ts
mykola-mokhnach: yes, this is internal wrapper to resolve/reject the wrapping promise, which also ensures the cleanup (timer + socket)

src/lib/tunnel/tunnel-availability.ts
mykola-mokhnach: I think it makes sense because this check is fast (less than 1ms) and allows to get a quick feedback before any actual tunnel request is made.

src/lib/tunnel/tunnel-availability.ts
mykola-mokhnach: this is done by the tunnel creation script, but I can also add it here for posterity

src/lib/tunnel/tunnel-availability.ts
mykola-mokhnach: I have updated the logic there to always throw TunnelAvailabilityError instead of a generic error. Wrapping is not needed anymore

harsha509 (approved):
Thank you @mykola-mokhnach !

PR #225, opened 2026-05-26
chore(tunnel): persist tunnel port to strongBox

Delta456:
This PR fixes a bug where the tunnel was connected to registry because it was never stored persistently. It caused new Mac devices not be able to connect to the tunnel despite running the creation script.

The fix is to store the tunnel port in a strongbox and write to that.

### DIFF
--- a/src/lib/tunnel/tunnel-registry-server.ts
+++ b/src/lib/tunnel/tunnel-registry-server.ts
@@ -1,5 +1,7 @@
+import { BaseItem, strongbox } from '@appium/strongbox';
 import * as http from 'node:http';
 
+import { TUNNEL_CONTAINER_NAME } from '../../constants.js';
 import { getLogger } from '../logger.js';
 import type { TunnelRegistry, TunnelRegistryEntry } from '../types.js';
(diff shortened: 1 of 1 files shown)


navin772 (approved):

Delta456:
This issue was found out while working on https://github.com/appium/appium-remote-debugger/issues/497

src/lib/tunnel/tunnel-registry-server.ts
```diff
@@ -391,6 +393,9 @@ export async function startTunnelRegistryServer(
 ): Promise<TunnelRegistryServer> {
   const server = new TunnelRegistryServer(tunnelInfos, port);
   await server.start();
+  const box = strongbox(TUNNEL_CONTAINER_NAME);
+  const item = new BaseItem('tunnelRegistryPort', box);
+  await item.write(String(port));
```
mykola-mokhnach: we should stop the server if this call throws an exception

Delta456: Good catch.

mykola-mokhnach (approved):

Issue #208, opened 2026-05-14
🐞 bug: iOS: Driver fails to start when RemoteXpcConnection started


amedvedjev:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Appium Version

3.4.2

### Appium Host Type

Command Line

### Is this issue reproducible using the latest components?

- [x] I confirm the issue is still reproducible with the latest component versions

### Current Behavior

Step 1: Start tunnel with sudo appium driver run xcuitest tunnel-creation --udid your_udid.
Step 2: Start appium driver. Mine test in fullReset option.

Result: Driver fails to start.

### Expected Behavior

Driver starting successfully. Without starting tunnel driver starting correctly.

### Appium Log

https://gist.github.com/amedvedjev/59c8b976be2e6983cd8959581f7b206e

### Environment Details

- Debug details (`appium --show-debug-info`): 

[appium_info.txt](https://github.com/user-attachments/files/27746373/appium_info.txt)

- Last component version(s) which did _not_ exhibit the problem: X
- Platform and version of device under test: iOS
- Real device or emulator/simulator: real iPhone11

### Minimal Reproducible Example

`driver[0] = new IOSDriver(new URL(baseURL + port + minorURL), optionsXCUITest);`

KazuCocoa:
What happens with `--disconnect-retry-max-attempts` for the tunnel command to re-connect when the client get disconnected.
https://appium.github.io/appium-xcuitest-driver/latest/reference/scripts/#tunnel-creation

amedvedjev:
to make it faster I set it to 1 and got last logs -> 
```
sudo appium driver run xcuitest tunnel-creation --udid 00008030-001D79403C84202E  --disconnect-retry-max-attempts 1


2026-05-14 07:43:26:386 - [ade4de54][XCUITestDriver@aab4] CFBundleSupportedPlatforms: ["iPhoneOS"]
2026-05-14 07:43:26:415 - [ade4de54][RemoteXpcConnection] Found 171 services and 74 ports
2026-05-14 07:43:26:416 - [ade4de54][RemoteXpcConnection] Extracted 171 services
2026-05-14 07:43:26:418 - [ade4de54][RemoteXpcConnection] Connection error: Error: read ECONNRESET
2026-05-14 07:43:26:418 - [ade4de54][RemoteXpcConnection] Socket closed
[...]

PR #227, opened 2026-05-30
fix: Add response demux and improve write framing for RSD AFC uploads

mykola-mokhnach:
## Summary

Fixes AFC install/upload hangs and timeouts when using RemoteXPC tunnels ([#208](https://github.com/appium/appium-ios-remotexpc/issues/208)).

The previous AFC client used a sequential send-then-receive model on a shared socket without routing responses by `packet_num`. That could desync under load (especially large IPA uploads with many 1 MiB WRITE chunks), leading to 15s operation timeouts and apparent driver hangs during `fullReset` app install.

This PR:

- Adds **`AfcPacketDemux`** — a background reader that routes inbound AFC packets to the matching request by `packet_num`, with serialized sends and per-operation timeouts.
- Fixes **`_connect()`** so the demux reader starts on connect and is reset only when the socket is replaced.
- Improves **WRITE packet framing** — three socket writes (header / handle / data) with correct `this_length` vs `entire_length`, avoiding per-chunk `Buffer.concat`.
- Enables **`TCP_NODELAY`** on AFC service sockets and uses **1 MiB** read/write chunks with a **30s** operation timeout.
- Removes **advisory FILE_LOCK** around uploads (not used on the reference install path; saves two round-trips per file).
- Adds a dedicated **AFC push performance integration test** for regression tracking on real devices/tunnels.
- Adds unit tests for demux routing and updated codec header tests.

**Note:** Benchmarking showed tunnel throughput also varies widely by stack (~58s vs ~8s vs ~1s for 10 MiB AFC over Appium tuntap / lockdown tunnel / go-ios tunnel). This PR fixes AFC protocol correctness and write efficiency; further install speed gains depend on **`appium-ios-tuntap`** (tracked separately).

## Changes

| Area | Files |
|---|---|
| Response demux | `src/services/ios/afc/demux.ts` (new) |
| AFC service | `src/services/ios/afc/index.ts` |
| Codec / writes | `src/services/ios/afc/codec.ts` |
| Constants | `src/services/ios/afc/constants.ts` |
| Streams | `src/services/ios/afc/stream-utils.ts` |
| Perf integration test | `test/integration/afc-push-performance-test.ts` (new) |
| Unit tests | `test/unit/afc/demux.spec.ts` (new), `test/unit/afc/codec.spec.ts` |
| npm script | `test:afc-push-perf` in `package.json` |
| Comment cleanup | `src/services/ios/testmanagerd/index.ts` |

## Integration test: AFC push performance

New test: `npm run test:afc-push-perf`

- Requires `UDID` and an active tunnel (`npm run test:tunnel-creation`)
- Uploads **10 MiB** (configurable via `AFC_PUSH_SIZE_BYTES`) to `/Downloads/…`
- Verifies remote file size via `stat`
- Asserts upload completes within **`AFC_PUSH_MAX_MS`** (default **120s**)
- Logs duration and **MiB/s** for manual/regression comparison

```bash
UDID=<udid> npm run test:afc-push-perf
UDID=<udid> AFC_PUSH_MAX_MS=60000 npm run test:afc-push-perf   # stricter budget after tuntap fix
```

## Test plan

- [ ] `npm run build`
- [ ] `npm run test:unit -- --grep 'AfcPacketDemux|AFC Codec'`
- [ ] `npm run test:unit` (full unit suite)
- [ ] With Appium tunnel running:
  ```bash
  UDID=<udid> npm run test:afc-push-perf
  ```
- [ ] `UDID=<udid> npm run test:afc` (existing AFC integration suite)
- [ ] Verify `listdir` / `stat` / `push` / `pull` over RSD AFC
- [ ] (Optional) Full 200 MiB IPA push via xcuitest-driver `fullReset` install with tunnel active — confirm no hang vs issue #208

### DIFF
--- a/src/services/ios/afc/codec.ts
+++ b/src/services/ios/afc/codec.ts
@@ -6,6 +6,7 @@ import {
   AFCMAGIC,
   AFC_HEADER_SIZE,
   AFC_OPERATION_TIMEOUT_MS,
+  MAXIMUM_READ_SIZE,
   NULL_BYTE,
 } from './constants.js';
 import { AfcError, AfcFopenMode, AfcOpcode } from './enums.js';
@@ -68,30 +69,33 @@ export function cstr(str: string): Buffer {
   return Buffer.concat([s, NULL_BYTE]);
 }
 /**
- * Build an AFC packet header for the provided operation and payload length.
+ * Build an AFC packet header with explicit entire_length and this_length values.
+ */
+export function encodeHeaderExplicit(
+  op: AfcOpcode,
+  packetNum: bigint,
+  entireLen: number,
+  thisLen: number,
+): Buffer {
+  const header = Buffer.alloc(AFC_HEADER_SIZE);
+  AFCMAGIC.copy(header, 0);
+  writeUInt64LE(BigInt(entireLen)).copy(header, 8);
+  writeUInt64LE(BigInt(thisLen)).copy(header, 16);
+  writeUInt64LE(packetNum).copy(header, 24);
+  writeUInt64LE(BigInt(op)).copy(header, 32);
+  return header;
+}
+
+/**
+ * Build an AFC packet header for a single contiguous payload after the header.
  */
 export function encodeHeader(
   op: AfcOpcode,
   packetNum: bigint,
   payloadLen: number,
-  thisLenOverride?: number,
 ): Buffer {
-  const entireLen = BigInt(AFC_HEADER_SIZE + payloadLen);
-  const thisLen = BigInt(thisLenOverride ?? AFC_HEADER_SIZE + payloadLen);
-
-  const header = Buffer.alloc(AFC_HEADER_SIZE);
-  // magic
-  AFCMAGIC.copy(header, 0);
-  // entire_length
-  writeUInt64LE(entireLen).copy(header, 8);
-  // this_length
-  writeUInt64LE(thisLen).copy(header, 16);
-  // packet_num
-  writeUInt64LE(packetNum).copy(header, 24);
-  // operation
-  writeUInt64LE(BigInt(op)).copy(header, 32);
-
-  return header;
+  const entireLen = AFC_HEADER_SIZE + payloadLen;
+  return encodeHeaderExplicit(op, packetNum, entireLen, entireLen);
 }
 
 const SOCKET_STATES = new WeakMap<net.Socket, SocketState>();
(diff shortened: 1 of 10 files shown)


src/services/ios/afc/codec.ts
```diff
@@ -201,33 +205,54 @@ export async function readAfcResponse(
...
+    const onDrain = () => {
+      cleanup();
+      resolve();
+    };
+    const onError = (err: Error) => {
+      cleanup();
+      reject(err);
+    };
+    const cleanup = () => {
+      socket.off('drain', onDrain);
```
harsha509, line 237: i see only drain and error. 
dont we need close/end ?

src/services/ios/afc/demux.ts
```diff
@@ -0,0 +1,215 @@
...
+  }
+
+  private _clearPending(pktNum: bigint): void {
+    this.pending.delete(pktNum);
+  }
+
+  private async _readerLoop(socket: net.Socket): Promise<void> {
+    try {
+      while (!this.stopped && !socket.destroyed) {
+        const header = await readAfcHeader(socket, AFC_OPERATION_TIMEOUT_MS);
```
harsha509, line 163: so, if there is no afc operation and idle for 30 seconds, socket is self-destroyed here

harsha509 (approved):

src/services/ios/afc/demux.ts
```diff
@@ -0,0 +1,215 @@
...
+    this.readerActive = false;
+  }
+
+  private async _runReaderLoop(socket: net.Socket): Promise<void> {
+    try {
+      await this._readerLoop(socket);
+    } catch (err) {
+      this._failPending(
+        err instanceof Error ? err : new Error(String(err)),
+        true,
```
harsha509, line 110: **nit:** a superseded reader has no `socket === this.readerSocket` guard, so it could notify fatal error for a replaced socket
low priority since the reconnect path is rarely reached and callers are sequential today.

src/services/ios/afc/demux.ts
mykola-mokhnach: yes, I think it's ok to avoid unexpected freezes

src/services/ios/afc/demux.ts
mykola-mokhnach: fixed

src/services/ios/afc/codec.ts
mykola-mokhnach: added throwing an error if the socket gets closed unexpectedly before drain

harsha509 (approved):
Lgtm

PR #234, opened 2026-06-10
feat: Only perform packet tap lazily

mykola-mokhnach:
## Summary

When the tunnel packet stream server is enabled, L4 packet tap was registered on the tunnel at startup — even with zero TCP clients connected. That forces every device→TUN frame through `tapL4Packet()` and adds ~7–15% overhead on hot paths like zip_conduit IPA installs.

This PR makes packet tap **lazy**: the tunnel consumer is attached only while at least one `PacketStreamServer` client is connected, and detached when the last client leaves or the server stops.

| Scenario | L4 parse on install path |
|----------|----------------------------|
| Packet stream server running, no clients | Off |
| Client connects | On |
| All clients disconnect | Off again |

### Changes

- **`PacketStreamServer`**: new `bindTunnel()` API; `start()` only opens the TCP listener; attach on first client, detach on last disconnect / `stop()`
- **`scripts/tunnel-creation.mjs`** and **`scripts/start-appletv-tunnel.mjs`**: call `bindTunnel(tunnel)` instead of `tunnel.addPacketConsumer(...)` at tunnel creation
- **Unit tests**: `test/unit/tunnel/packet-stream-server.spec.ts` covers attach/detach lifecycle

### Motivation

Benchmarks on a ~200 MiB IPA install:

| Path | Install time |
|------|-------------|
| `appium-ios-device` (no tunnel) | ~14s |
| remotexpc zip_conduit, packet tap off | ~20–21s |
| remotexpc zip_conduit, packet tap on | ~22–25s |

With lazy attach, running tunnel-creation with packet stream enabled but no connected clients should match tap-off performance. Tap cost is only paid when something is actively capturing packets.

### DIFF
--- a/src/lib/tunnel/packet-stream-server.ts
+++ b/src/lib/tunnel/packet-stream-server.ts
@@ -3,8 +3,14 @@ import { EventEmitter } from 'node:events';
 import { type Server, type Socket, createServer } from 'node:net';
 
 import { getLogger } from '../logger.js';
+import type { PacketSource } from '../types.js';
 
 const log = getLogger('PacketStreamServer');
+
+type TunnelPacketBinding = Pick<
+  PacketSource,
+  'addPacketConsumer' | 'removePacketConsumer'
+>;
 
 /**
  * Server that exposes packet streaming from a tunnel over TCP
@@ -14,9 +20,21 @@ export class PacketStreamServer extends EventEmitter {
   private server: Server | null = null;
   private readonly clients: Set<Socket> = new Set();
   private packetConsumer: PacketConsumer | null = null;
+  private tunnel: TunnelPacketBinding | null = null;
+  private attachedToTunnel = false;
 
   constructor(private readonly port: number) {
     super();
+  }
+
+  /**
+   * Bind the tunnel whose packet consumer is attached only while TCP clients are connected.
+   */
+  bindTunnel(tunnel: TunnelPacketBinding): void {
+    this.tunnel = tunnel;
+    if (this.clients.size > 0) {
+      this.attachToTunnel();
+    }
   }
 
   /**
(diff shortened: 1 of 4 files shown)


test/unit/tunnel/packet-stream-server.spec.ts
```diff
@@ -0,0 +1,171 @@
...
+    server.bindTunnel(tunnel);
+    await server.start();
+
+    const client1 = await connectClient(port);
+    const client2 = await connectClient(port);
+
+    expect(added).to.have.lengthOf(1);
+
+    await closeClient(client1);
+    await waitFor(() => removed.length === 0);
```
harsha509: this is no-op, may predicate true immediately

src/lib/tunnel/packet-stream-server.ts
```diff
@@ -14,11 +20,23 @@ export class PacketStreamServer extends EventEmitter {
...
+  private attachedToTunnel = false;
 
   constructor(private readonly port: number) {
     super();
   }
 
+  /**
+   * Bind the tunnel whose packet consumer is attached only while TCP clients are connected.
+   */
+  bindTunnel(tunnel: TunnelPacketBinding): void {
```
harsha509, line 33: do we need to do a cleanup? may be `this.detachFromTunnel()` as first step
say `bindTunnel(tunnelA)` connects to A
when i say bindTunnel(tunnelB) , it never clears A

mykola-mokhnach: the plan was to only support single tunnel attachment. Would you like to update it, so binding to another tunnel overrides the previous one? It may also be confusing if the server is already serving clients

test/unit/tunnel/packet-stream-server.spec.ts
mykola-mokhnach: fixed

src/lib/tunnel/packet-stream-server.ts
harsha509: oh then single attachment is fine i think.... we can leave as it is

harsha509 (approved):
Thank you @mykola-mokhnach !

PR #237, opened 2026-06-14
feat: Drop tunnel packets tap

mykola-mokhnach:
## Summary

Binary `os_trace_relay` syslog no longer depends on tunnel packet tap. It reads framed entries directly from the RSD service TCP socket after `StartActivity`, matching pymobiledevice3 and the existing text syslog path.

The packet tap stack (`PacketStreamServer`, `PacketStreamClient`, registry `packetStreamPort`) is removed entirely — it was only used for binary syslog.

**~1,000 lines removed**, with a focused socket-based syslog implementation.

## Motivation

Packet tap was an remotexpc-specific workaround: binary syslog sniffed L4 payloads from the tunnel forwarder via a cross-process TCP bridge. Apple’s model use a normal RSD service connection — handshake on the socket, then read the binary stream on that same connection.

Removing tap also means `TunnelManager.hasPacketTap()` stays false during normal operation, so tuntap always uses the lighter `DeviceToTunPump` / `TunToDevicePump` path instead of the buffer + tap forwarding path.

### SyslogService (binary mode)
- Send `StartActivity` via plist, then `socket.unpipe()` and feed raw bytes into `SyslogProtocolParser` (same pattern as text mode).
- **`SyslogService.start(service, options?)`** — `packetSource` parameter removed.

### Removed
- `PacketStreamServer` / `PacketStreamClient` and all tunnel daemon wiring
- `packetStreamPort` from tunnel registry entries and `TunnelEndpoint`
- `PacketSource` / `PacketConsumer` types and public exports
- `--packet-stream-base-port` CLI options from tunnel scripts

### Tests
- Updated `test/integration/tunnel-test.ts` — no `PacketStreamClient`
- Added `test/unit/syslog/syslog-service.spec.ts` — socket framing unit test

## Breaking changes

- **BREAKING CHANGE:** `SyslogService.start()` no longer accepts a `packetSource` argument. Call `syslogService.start(descriptor, options)` instead; binary syslog reads from the RSD service socket after `StartActivity`.

- **BREAKING CHANGE:** Tunnel registry entries and `TunnelEndpoint` no longer expose `packetStreamPort`.

- **BREAKING CHANGE:** `PacketStreamServer` and `PacketStreamClient` are removed from the public API.

Downstream callers (e.g. Appium) must stop passing a packet source and stop using the packet stream bridge.


test/unit/syslog/syslog-service.spec.ts
```diff
@@ -0,0 +1,88 @@
...
+  header.writeUInt8(0x02, 0);
+  header.writeUInt32LE(entry.length, 1);
+  return Buffer.concat([header, entry]);
+}
+
+describe('SyslogService binary mode', function () {
+  it('reads framed syslog entries from the service socket after StartActivity', async function () {
+    const socket = new PassThrough();
+    const fakeConnection = {
+      sendPlistRequest: async () => ({ Status: 'RequestSuccessful' }),
```
harsha509, line 46: i see `{ Request: 'StartService' }` instead of `{ Status: 'RequestSuccessful' }`.

harsha509 (approved):
Thank you @mykola-mokhnach !

test/unit/syslog/syslog-service.spec.ts
mykola-mokhnach: { Request: 'StartService' } is the greeting you get during the RSD check-in handshake (after RSDCheckin). Other services validate that explicitly, but syslog does not on this call — and in this test that handshake is skipped anyway because BaseService is mocked to return fakeConnection directly.

So:

{ Status: 'RequestSuccessful' } — correct for the StartActivity response this test exercises
{ Request: 'StartService' } — wrong shape for this call; it would only “work” because Status is undefined, so the failure check is skipped, but it does not match real behavior
The test name (“after StartActivity”) and the assertion path both align with keeping { Status: 'RequestSuccessful' }.

PR #241, opened 2026-06-19
fix: lint warnings

mykola-mokhnach:


KazuCocoa (approved):

src/services/ios/syslog-service/syslog-entry-parser.ts
```diff
@@ -171,7 +171,7 @@ export function parseSyslogEntry(data: Buffer): SyslogEntry {
       categorySize > 1
         ? decodeUtf8(data.subarray(offset, offset + categorySize - 1))
         : '';
-    offset += categorySize;
+    // offset += categorySize;
```
KazuCocoa, line 174: btw, because of no usage?
Then, can be removed instead of comment out?

mykola-mokhnach: yes, I decided to keep it as comment for posterity, in case someone would decide to add something on top of it.

PR #245, opened 2026-06-20
chore: Improve progress logging for AppleTV pairing scenarios

mykola-mokhnach:
I would like to add similar changes to the pairing script as I did to the tunnel creation one in https://github.com/appium/appium-xcuitest-driver/pull/2879

### DIFF
--- a/scripts/pair-appletv.mjs
+++ b/scripts/pair-appletv.mjs
@@ -8,9 +8,98 @@
 
 import { logger } from '@appium/support';
 import { Command } from 'commander';
-import { AppleTVPairingService, UserInputService } from 'appium-ios-remotexpc';
+import {
+  AppleTVPairingService,
+  UserInputService,
+} from 'appium-ios-remotexpc';
+import { DEFAULT_PAIRING_CONFIG } from '../build/src/lib/apple-tv/constants.js';
 
 const log = logger.getLogger('AppleTVPairing');
+const APPLETV_PAIRING_DISCOVERY_PROGRESS_INTERVAL_MS = 1000;
+const APPLETV_PAIRING_DISCOVERY_PROGRESS_BAR_WIDTH = 24;
+
+function parsePositiveInteger(value) {
+  const count = Number.parseInt(value, 10);
+  if (!Number.isFinite(count) || count <= 0) {
+    throw new Error(
+      `Invalid timeout: ${value}. Expected a positive integer in milliseconds.`,
+    );
+  }
+  return count;
+}
+
+function startTimeoutProgressLogger({
+  label,
+  startedAt,
+  timeoutMs,
+  barWidth,
+  intervalMs,
+}) {
+  let timer = null;
+  let isStopped = false;
+
+  const logProgress = (status, isComplete = false) => {
+    const elapsedMs = performance.now() - startedAt;
+    const boundedElapsedMs = Math.min(elapsedMs, timeoutMs);
+    const progress = isComplete ? 1 : boundedElapsedMs / timeoutMs;
+    const filledWidth = Math.round(progress * barWidth);
+    const emptyWidth = barWidth - filledWidth;
+    const bar = `${'#'.repeat(filledWidth)}${'-'.repeat(emptyWidth)}`;
+    log.info(
+      `${label}: [${bar}]${status && status !== 'waiting' ? ` - ${status}` : ''}`,
+    );
+  };
+
+  const stop = (status, isComplete = false) => {
+    if (isStopped) {
+      return;
+    }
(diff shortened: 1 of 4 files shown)


KazuCocoa (approved):

scripts/pair-appletv.mjs
```diff
@@ -8,9 +8,98 @@
...
+  UserInputService,
+} from 'appium-ios-remotexpc';
+import { DEFAULT_PAIRING_CONFIG } from '../build/src/lib/apple-tv/constants.js';
 
 const log = logger.getLogger('AppleTVPairing');
+const APPLETV_PAIRING_DISCOVERY_PROGRESS_INTERVAL_MS = 1000;
+const APPLETV_PAIRING_DISCOVERY_PROGRESS_BAR_WIDTH = 24;
+
+function parsePositiveInteger(value) {
+  const count = Number.parseInt(value, 10);
```
harsha509, line 22: may be Number(value) to reject typos ??

scripts/pair-appletv.mjs
```diff
@@ -8,9 +8,98 @@
...
+
+  const logProgress = (status, isComplete = false) => {
+    const elapsedMs = performance.now() - startedAt;
+    const boundedElapsedMs = Math.min(elapsedMs, timeoutMs);
+    const progress = isComplete ? 1 : boundedElapsedMs / timeoutMs;
+    const filledWidth = Math.round(progress * barWidth);
+    const emptyWidth = barWidth - filledWidth;
+    const bar = `${'#'.repeat(filledWidth)}${'-'.repeat(emptyWidth)}`;
+    log.info(
+      `${label}: [${bar}]${status && status !== 'waiting' ? ` - ${status}` : ''}`,
```
harsha509, line 49: The progress bar fills up based on the discovery timeout (e.g. 10 seconds). but after that, the code does one more step, fetching extra device info with `devicectl` and the bar keeps printing as full [########] every second during that step. 

It looks frozen/stuck even though work is still happening.

scripts/pair-appletv.mjs
```diff
@@ -8,9 +8,98 @@
...
+  }, intervalMs);
+  timer.unref?.();
+
+  return {
+    succeed: (message = 'done') => stop(message, true),
+    fail: (message = 'failed') => stop(message),
+  };
+}
+
+function discoverAppleTVPairingDevices(pairingService, timeoutMs) {
```
harsha509, line 77: these two are called back to back

`async function waitForAppleTVPairingDiscovery(discovery, timeoutMs) { ... }`

any idea of mergingi them ?

scripts/pair-appletv.mjs
```diff
@@ -8,9 +8,98 @@
 
 import { logger } from '@appium/support';
 import { Command } from 'commander';
-import { AppleTVPairingService, UserInputService } from 'appium-ios-remotexpc';
+import {
+  AppleTVPairingService,
+  UserInputService,
+} from 'appium-ios-remotexpc';
+import { DEFAULT_PAIRING_CONFIG } from '../build/src/lib/apple-tv/constants.js';
```
harsha509, line 15: reading from build ?

src/lib/apple-tv/pairing/pairing-service.ts
```diff
@@ -61,7 +74,6 @@ export class AppleTVPairingService {
         pairingFile,
       };
     } catch (error) {
-      log.error('Pairing failed:', error);
```
harsha509, line 64: removing this users might see raw node dumps on failed pairing i think

mykola-mokhnach: not sure what you mean, do you have any real example to compare?

scripts/pair-appletv.mjs
mykola-mokhnach: yes, this is fine. In the deployed package build will always be there. In the dev one may build it whenever neeeded

scripts/pair-appletv.mjs
mykola-mokhnach: does the latter takes so long? I thought only bonjour discovery is the longest op and needs some progress to improve the user feedback.

scripts/pair-appletv.mjs
mykola-mokhnach: I don't have any strong opinion on that. Would you prefer it to be more strict?

scripts/pair-appletv.mjs
mykola-mokhnach: this should look better in https://github.com/appium/appium-ios-remotexpc/pull/247

Issue #159, opened 2026-03-15
Use Apple TV serial number as tunnel registry UDID when available


mykola-mokhnach:
Today the tunnel registry uses the Bonjour identifier (from _remotepairing-manual-pairing TXT) as the Apple TV UDID. For consistency with USB iOS devices (which use SerialNumber as the registry key), we should use the Apple TV serial number as the tunnel UDID when we can obtain it, and fall back to the Bonjour identifier otherwise.

Proposed approach

1. Pairing: In the M6 response handler, read TLV type SERIAL_NUMBER (0x17) from the decrypted payload when present. Extend the pair record (and plist storage) with an optional serialNumber and persist it at pairing time.
2. Tunnel creation: When starting the Apple TV tunnel, load the pair record by Bonjour identifier; if it has serialNumber, return it from the tunnel service. In start-appletv-tunnel.ts, set the registry entry’s udid to serialNumber ?? deviceInfo.identifier.
3. Compatibility: Existing pair records without serial_number keep using deviceInfo.identifier; new pairings that receive 0x17 in M6 will use the serial as UDID.

Note on reliability

pymobiledevice3 and pyatv do not document or use Apple TV hardware serial over WiFi. There is no public confirmation that Apple TV sends SERIAL_NUMBER (0x17) in M6; it is the only in-protocol candidate in our flow. Implementation is low cost; behavior should be confirmed on real hardware and documented as “serial when provided, otherwise Bonjour identifier.”

Relevant code

- M6 handling: src/lib/apple-tv/pairing-protocol/pairing-protocol.ts (receiveM6Completion)
- Pair record type/storage: src/lib/apple-tv/storage/types.ts, src/lib/apple-tv/storage/pairing-storage.ts
- Tunnel script registry: scripts/start-appletv-tunnel.ts
- TLV constant: SERIAL_NUMBER: 0x17 in src/lib/apple-tv/constants.ts

mykola-mokhnach:
Closed as fixed

PR #246, opened 2026-06-20
fix: Fetch AppleTV device udid from M6 pairing record

mykola-mokhnach:
## Summary
- Decode the decrypted Apple TV M6 `INFO` OPACK2 payload and extract `remotepairing_udid`.
- Persist `remote_pairing_udid` in Apple TV pair records and use it as the primary tunnel registry UDID.
- Keep macOS `devicectl` identifiers as a fallback only when `remote_pairing_udid` is unavailable, and fail otherwise.
- Normalize Apple TV UDIDs to uppercase for consistency with `devicectl` and registry expectations.

Addresses https://github.com/appium/appium-ios-remotexpc/issues/159

### DIFF
--- a/src/lib/apple-tv/devicectl-enrichment.ts
+++ b/src/lib/apple-tv/devicectl-enrichment.ts
@@ -12,7 +12,12 @@ export async function enrichDiscoveredDevicesWithDevicectl(
     return devices;
   }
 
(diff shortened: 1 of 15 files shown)


harsha509 (approved):
Thank you @mykola-mokhnach !

PR #247, opened 2026-06-20
chore: refactor tunnels creation and pairing scripts

mykola-mokhnach:
## Summary

- Refactor Apple TV pairing and tunnel helper scripts to share common code through `scripts/lib`.
- Move repeated option parsing, progress logging, timers, root/admin checks, registry defaults, and tunnel registry publishing helpers into reusable script utilities.
- Keep existing CLI flags compatible while improving no-device handling and pluralized log output.


scripts/tunnel-creation.mjs
```diff
@@ -419,6 +364,7 @@ async function createTunnelForDevice(device) {
 }
 
 async function main() {
+  await assertRoot('tunnel-creation');
```
harsha509: can we make --help without root ? or root is required

as i see `assertRoot()` runs before `program.parse()`, so `--help` now requires root

harsha509 (approved):
Thank you @mykola-mokhnach !
LGTM!

navin772 (commented):
Currently, when running the appletv tunnel script when no tvos devices are connected it hangs and does not closes the HTTP server.
I think we should close it just like we do for iOS devices.

navin772 (approved):
LGTM!

scripts/tunnel-creation.mjs
mykola-mokhnach: updated

mykola-mokhnach:
> Currently, when running the appletv tunnel script when no tvos devices are connected it hangs and does not closes the HTTP server. I think we should close it just like we do for iOS devices.

Fixed now

PR #251, opened 2026-06-23
feat(sysmontap): add sysmontap RSD service

navin772:
Adds sysmontap RSD service - `com.apple.instruments.server.services.sysmontap` to retrieve any set of metrics on a specific interval.

Related to #101 

Usage:

```typescript
const { sysmontap, dvtService } = await Services.startDVTService(udid);
try {
  for await (const procs of sysmontap.iterProcesses()) {
    const app = procs.find(p => p.name === 'MyApp');
    console.log(app?.cpuUsage, app?.physFootprint);
    break; // or keep streaming
  }
} finally {
  await dvtService.close();
}
```


src/services/ios/dvt/instruments/sysmontap.ts
```diff
@@ -0,0 +1,268 @@
...
+    );
+
+    this.builtConfig = {
+      ur: Sysmontap.MINIMUM_INTERVAL_MS, // Output frequency (ms)
+      bm: 0,
+      procAttrs: this.processAttributes,
+      sysAttrs: this.systemAttributes,
+      cpuUsage: true,
+      physFootprint: true, // Include physical memory footprint
+      sampleInterval: intervalMs * 1_000_000, // Sample interval in nanoseconds
```
mykola-mokhnach, line 105: would it make sense to reduce interval precision to microseconds instead of milliseconds?

src/services/ios/dvt/instruments/sysmontap.ts
```diff
@@ -0,0 +1,268 @@
...
+   * A sysmontap instance supports a single sampling session per DVT
+   * connection: once a stream has been {@link stop}ped, the device does not
+   * resume it on the same connection. To sample again, start a new DVT
+   * connection (e.g. via `Services.startDVTService`).
+   */
+  async start(): Promise<void> {
+    if (!this.builtConfig) {
+      await this.configure();
+    }
+    await this.initialize();
```
mykola-mokhnach, line 132: should we allow to call start twice is already started?

src/services/ios/dvt/instruments/sysmontap.ts
```diff
@@ -0,0 +1,268 @@
...
+  /**
+   * Async iterator that yields raw sysmontap samples as they arrive. The
+   * device interleaves two kinds of samples: system samples (with `System`,
+   * `SystemAttributes`, CPU usage) and process samples (with `Processes`).
+   * Internal control/heartbeat frames are filtered out.
+   *
+   * Sampling starts automatically on first iteration and stops when iteration
+   * terminates (via break, return, or error).
+   */
+  async *messages(): AsyncGenerator<SysmonSample, void, unknown> {
```
mykola-mokhnach, line 183: does it stop automatically if dvt is stopped?

src/services/ios/dvt/instruments/sysmontap.ts
```diff
@@ -0,0 +1,268 @@
...
+        this.receiveAbortController = receiveAbortController;
+
+        let plist: unknown;
+        try {
+          plist = await channel.receivePlist(receiveAbortController.signal);
+        } catch (err) {
+          if (this.stopRequested && this.isAbortError(err)) {
+            break;
+          }
+          throw err;
```
mykola-mokhnach: I am not sure if we ever want to throw an exception from this loop as this would result in an unhandled exception

src/services/ios/dvt/instruments/sysmontap.ts
```diff
@@ -0,0 +1,268 @@
...
+    }
+  }
+
+  /**
+   * Map a raw sample's `System` tuple to a labelled object using the configured
+   * system attribute names.
+   * @param sample A sample yielded by {@link messages}
+   * @returns The labelled system info, or `null` when the sample has no system data
+   */
+  parseSystem(sample: SysmonSample): SysmonSystemInfo | null {
```
mykola-mokhnach: should this method be private?

mykola-mokhnach: not sure why it's needed in general, it does not do much

src/services/ios/dvt/instruments/sysmontap.ts
navin772: not required since we emit only ~2 samples/sec

src/services/ios/dvt/instruments/sysmontap.ts
navin772: will add an idempotency guard

src/services/ios/dvt/instruments/sysmontap.ts
navin772: not currently, will update to end stream on connection close/failure

src/services/ios/dvt/instruments/sysmontap.ts
navin772: will fix this

src/services/ios/dvt/instruments/sysmontap.ts
navin772: i think we can have a public `iterSystem()` generator similar to `iterProcesses()` instead of this standalone helper and make `parseSystem` private

src/services/ios/dvt/instruments/sysmontap.ts
```diff
@@ -0,0 +1,312 @@
...
+      procAttrs: this.processAttributes,
+      sysAttrs: this.systemAttributes,
+      cpuUsage: true,
+      physFootprint: true, // Include physical memory footprint
+      sampleInterval: intervalMs * 1_000_000, // Sample interval in nanoseconds
+    };
+
+    log.debug(
+      `sysmontap configured: interval=${intervalMs}ms, ` +
+        `${this.processAttributes.length} process attrs, ` +
```
mykola-mokhnach: use util.pluralize to properly reflect plurals

src/services/ios/dvt/instruments/sysmontap.ts
```diff
@@ -0,0 +1,312 @@
...
+        await this.requireChannel().call('stop')(undefined, false);
+      } catch (error) {
+        log.debug(
+          'sysmontap stop() could not notify the device:',
+          error instanceof Error ? error.message : error,
+        );
+      }
+    }
+    this.started = false;
+    log.debug('sysmontap sampling stopped');
```
mykola-mokhnach, line 168: this log would be confusing if the sysmon has been already stopped/was not running

src/services/ios/dvt/instruments/sysmontap.ts
```diff
@@ -0,0 +1,312 @@
...
+        yield system;
+      }
+    }
+  }
+
+  /**
+   * Map a raw sample's `System` tuple to a labelled object using the configured
+   * system attribute names, or `null` when the sample has no system data.
+   */
+  private parseSystem(sample: SysmonSample): SysmonSystemInfo | null {
```
mykola-mokhnach: not sure if this method makes or we could just inline this logic similarly to iterProcesses?

mykola-mokhnach (approved):

src/services/ios/dvt/instruments/sysmontap.ts
navin772: we can inline it

src/services/ios/dvt/instruments/sysmontap.ts
navin772: will log only when sysmon is started

Delta456 (approved):

harsha509 (approved):
LGTM!

PR #252, opened 2026-06-23
feat(coredevice): add coredevice app-service

navin772:
Adds the `com.apple.coredevice.appservice` RemoteXPC service and the shared infrastructure to build `coredevice.*` request/response services.

### What's included
- **`CoreDeviceService` base** (`src/services/ios/core-device/`): wraps requests in the CoreDevice invocation envelope, adds `invoke()` (request/response) and`send()` (fire-and-forget), and opens a fresh connection per `invoke()` since
  CoreDevice services are one-shot per connection.
- **`AppServiceService`** (`src/services/ios/app-service/`): `launchApplication`, `terminateApplication`, `sendSignalToProcess`, `listProcesses`, `uninstallApp`, `listApps`, `monitorProcessTermination`. Exposed via `Services.startAppServiceService(udid)`.
- **HID migrated** onto the shared base (removes duplicated transport lifecycle).
- **HTTP/2 fix:** initial receive window raised 1 MiB → 16 MiB so large RemoteXPC replies don't stall on the (non-replenished) odd reply channel.

pymobiledevice3 implementation - https://github.com/doronz88/pymobiledevice3/blob/master/pymobiledevice3/remote/core_device/


src/services/ios/app-service/index.ts
```diff
@@ -0,0 +1,280 @@
...
+  XPCDictionary,
+  XPCValue,
+} from '../../../lib/types.js';
+import {
+  type CoreDeviceInvokeOptions,
+  CoreDeviceService,
+} from '../core-device/core-device-service.js';
+
+/** POSIX signal numbers commonly used to terminate processes. */
+export const SIGNAL_SIGTERM = 15;
```
mykola-mokhnach: consider using standard node constants:

```
import { constants } from 'node:os';

console.log(constants.signals.SIGINT);
```

src/services/ios/app-service/index.ts
```diff
@@ -0,0 +1,280 @@
...
+        includeAppGroupIdentifiers: options.includeAppGroupIdentifiers ?? false,
+        includeContainerPaths: options.includeContainerPaths ?? false,
+      },
+      { timeoutMs: options.timeoutMs },
+    );
+    return asArray(output) as InstalledApp[];
+  }
+
+  /**
+   * Launches an application by bundle identifier and returns its process token.
```
mykola-mokhnach, line 159: what happens if app with given bundle does not exist?

src/services/ios/app-service/index.ts
```diff
@@ -0,0 +1,280 @@
...
+  /**
+   * Lists the processes currently running on the device.
+   */
+  async listProcesses(): Promise<AppServiceProcessToken[]> {
+    const output = asDict(await this.invoke(FEATURE_LIST_PROCESSES));
+    return asArray(output.processTokens) as AppServiceProcessToken[];
+  }
+
+  /**
+   * Sends a POSIX signal to a process by pid.
```
mykola-mokhnach: what happens if pid does not exist?

src/services/ios/app-service/index.ts
```diff
@@ -0,0 +1,280 @@
...
+        process: { processIdentifier: pid },
+        signal,
+      }),
+    );
+  }
+
+  /**
+   * Terminates a process by pid (sends SIGKILL by default).
+   */
+  async terminateApplication(
```
mykola-mokhnach: not sure we need this wrapper, it does not do much

src/services/ios/app-service/index.ts
```diff
@@ -0,0 +1,280 @@
...
+    pid: number,
+    signal: number = SIGNAL_SIGKILL,
+  ): Promise<XPCDictionary> {
+    return this.sendSignalToProcess(pid, signal);
+  }
+
+  /**
+   * Uninstalls an application by bundle identifier.
```
mykola-mokhnach: what happens if the app is already uninstalled?

src/services/ios/app-service/index.ts
```diff
@@ -0,0 +1,280 @@
...
+   */
+  async uninstallApp(bundleId: string): Promise<void> {
+    await this.invoke(FEATURE_UNINSTALL_APP, { bundleIdentifier: bundleId });
+  }
+
+  /**
+   * Monitors termination of a process. Resolves when the device reports the
```
mykola-mokhnach: what if the pid is already not running when this api is called?
should we also be able to provide a timeout?

src/services.ts
```diff
@@ -123,6 +124,19 @@ export async function startHidIndigoService(
   return new HidIndigoService(udid);
 }
 
+/**
+ * Start the CoreDevice AppService for the given device UDID.
+ *
+ * Provides app and process management (list apps, launch/terminate apps, list
+ * processes, signal processes, uninstall apps) over RemoteXPC.
+ */
+export async function startAppServiceService(
```
mykola-mokhnach: can we have a better method name so Service is not repeated twice?

src/services/ios/core-device/core-device-service.ts
```diff
@@ -0,0 +1,303 @@
...
+  readonly response?: XPCDictionary;
+
+  constructor(message: string, response?: XPCDictionary) {
+    super(message);
+    this.name = 'CoreDeviceError';
+    this.response = response;
+  }
+}
+
+export interface CoreDeviceInvokeOptions {
```
mykola-mokhnach: Please fix linter issues

src/services/ios/core-device/core-device-service.ts
```diff
@@ -0,0 +1,303 @@
...
+      this.nextMessageId = 1;
+    }
+    return this.transport;
+  }
+
+  /**
+   * Closes any existing connection and opens a fresh one. Used by {@link invoke}
+   * because CoreDevice services are one-shot per connection.
+   */
+  private async freshTransport(): Promise<RemoteXpcFramedTransport> {
```
mykola-mokhnach: maybe refreshTransport?

src/services/ios/app-service/index.ts
navin772: will remove it

src/services/ios/app-service/index.ts
navin772: it is idempotent, will document it

harsha509 (approved):
LGTM!

src/services.ts
navin772: lets just do `startAppService` and `AppService`

src/services/ios/app-service/index.ts
navin772: it resolves immediately, timeout can be provided via options:
```
await appService!.monitorProcessTermination(987654, {
      timeoutMs: 8000,
    });
```

src/services/ios/app-service/index.ts
navin772: throws a `CoreDeviceError`:
`CoreDeviceError: CoreDevice 'com.apple.coredevice.feature.launchapplication' failed: The application failed to launch. The requested application com.foo.doesnotexist is not installed.`

src/services/ios/app-service/index.ts
navin772: throws `CoreDeviceError` :
```
CoreDeviceError: CoreDevice 'com.apple.coredevice.feature.sendsignaltoprocess' failed: Failed to send signal 9 to process 1,394. No such process; the process may have already terminated. [com.apple.dt.CoreDeviceError 10014]
```

src/services/ios/app-service/index.ts
mykola-mokhnach: Please mention it in jsdoc

src/services/ios/app-service/index.ts
mykola-mokhnach: Please mention it in jsdoc

src/services/ios/app-service/index.ts
mykola-mokhnach: we can keep it if we add a possibility to map app bundle id to a pid

src/services/ios/app-service/index.ts
```diff
@@ -0,0 +1,280 @@
...
+    signal: number,
+  ): Promise<XPCDictionary> {
+    return asDict(
+      await this.invoke(FEATURE_SEND_SIGNAL, {
+        process: { processIdentifier: pid },
+        signal,
+      }),
+    );
+  }
+
```
mykola-mokhnach, line 229: can we add a method to check if an app is installed?

src/services/ios/core-device/core-device-service.ts
```diff
@@ -0,0 +1,351 @@
...
+      return;
+    }
+
+    const transport = this.transport;
+    this.transport = null;
+    await transport.close();
+  }
+
+  /**
+   * Sends a fire-and-forget XPC message on the root channel. Used by services
```
mykola-mokhnach, line 97: do we properly handle possible exceptions on FaF messages, so there is no potential possibility for unhandled ones?

mykola-mokhnach (approved):

src/services/ios/app-service/index.ts
navin772: To check this, we would need listApps from coredevice, but currently listApps is not working as intended and does not returns anything (documented):
```
Failed to invoke: com.apple.coredevice.feature.listapps. Got error: {'CoreDevice.error': {'userInfoWithNSSecureCoding': ....
```

It is something from Apple's end, they are not allowing to retrieve the list with both `requireContainerAccess:false` and `true`.

src/services/ios/core-device/core-device-service.ts
navin772: fixed

src/services/ios/app-service/index.ts
navin772: it will depend on listApps which is not reliable, can be implemented later when this behaviour is fixed in app-service

PR #254, opened 2026-06-25
feat(dvt): implement energy monitor

Delta456:
This PR implements energy monitor DVT service (`com.apple.xcode.debug-gauge-data-providers.Energy`). Implementation referred from https://github.com/doronz88/pymobiledevice3/blob/master/pymobiledevice3/services/dvt/instruments/energy_monitor.py


src/services/ios/dvt/instruments/energy-monitor.ts
```diff
@@ -0,0 +1,90 @@
...
+ * for await (const sample of energy.monitor([1234])) {
+ *   console.log(sample);
+ * }
+ * ```
+ */
+export class EnergyMonitor extends BaseInstrument {
+  static readonly IDENTIFIER =
+    'com.apple.xcode.debug-gauge-data-providers.Energy';
+
+  async startSampling(pids: number[]): Promise<void> {
```
mykola-mokhnach, line 51: what happens if this method is invoked twice in a row?

src/services/ios/dvt/instruments/energy-monitor.ts
```diff
@@ -0,0 +1,90 @@
...
+    await this.initialize();
+    const channel = this.requireChannel();
+    await channel.call('startSamplingForPIDs_')(
+      new MessageAux().appendObj(pids),
+      false,
+    );
+  }
+
+  async stopSampling(pids: number[]): Promise<void> {
```
mykola-mokhnach, line 70: what happens if this method is invoked twice in a row?

src/lib/types.ts
```diff
@@ -815,6 +815,39 @@ export interface NetworkMonitorService {
...
+  startSampling(pids: number[]): Promise<void>;
+  /**
+   * Stop energy sampling for the given PIDs.
+   */
+  stopSampling(pids: number[]): Promise<void>;
+  /**
+   * Take a single energy snapshot for the given PIDs.
+   * Returns a record mapping PID (string key) to its energy metrics.
+   */
+  sample(pids: number[]): Promise<Record<string, Record<string, number>>>;
```
mykola-mokhnach, line 835: is it required to call startSampling to invoke this method? If not then what's the practical purpose of above APIs?

src/services/ios/dvt/instruments/energy-monitor.ts
```diff
@@ -0,0 +1,90 @@
...
+   */
+  async *monitor(
+    pids: number[],
+  ): AsyncGenerator<EnergyMonitorSample, void, undefined> {
+    log.debug('Energy monitoring started');
+    // initialize() before stopSampling so requireChannel() doesn't throw on first use
+    await this.initialize();
+    // Stop first in case a previous session is still active
+    await this.stopSampling(pids);
+    await this.startSampling(pids);
```
mykola-mokhnach, line 118: I expect startSampling to be idempotent (e.g. further call should be ignored if sampling is already running)

src/services/ios/dvt/instruments/energy-monitor.ts
```diff
@@ -0,0 +1,90 @@
...
+
+    try {
+      while (true) {
```
mykola-mokhnach: we need to make sure this loop stops if either dvt service is closed or sampling is stopped

src/services/ios/dvt/index.ts
```diff
@@ -225,28 +225,28 @@ export class DVTSecureSocketProxyService extends BaseService {
     const payloadHeader = DTXMessage.parsePayloadHeader(packetData);
 
     const compression = (payloadHeader.flags & 0xff000) >> 12;
-    if (compression) {
-      throw new Error('Compressed messages not supported');
+
+    if (compression !== 0) {
```
mykola-mokhnach, line 229: are these updates necessary for this feature or this is a separate fix?

src/services/ios/dvt/instruments/energy-monitor.ts
```diff
@@ -0,0 +1,90 @@
...
+    pids: number[],
+  ): AsyncGenerator<EnergyMonitorSample, void, undefined> {
+    log.debug('Energy monitoring started');
+    // initialize() before stopSampling so requireChannel() doesn't throw on first use
+    await this.initialize();
+    // Stop first in case a previous session is still active
+    await this.stopSampling(pids);
+    await this.startSampling(pids);
+
+    try {
```
mykola-mokhnach, line 120: what is the execution frequency there. May yielding it without a delay create excessive CPU usage?

src/services/ios/dvt/index.ts
Delta456: Yes, `DTXBlockCompression: 2` change is required for this feature as it tells device to send LZ4-compressed responses which includes real time energy metrics

src/services/ios/dvt/instruments/energy-monitor.ts
Delta456: Right. Will see how it has to be done.

src/services/ios/dvt/instruments/energy-monitor.ts
Delta456: Right

src/lib/types.ts
Delta456: Yes, it is required as device hasn't begun its collection cycle and `sampleAttributes:forPIDs:` returns empty metrics. Will add this to docs.

src/services/ios/dvt/instruments/energy-monitor.ts
Delta456: It should silently be ignored but may cause issues. I will check and update

src/services/ios/dvt/instruments/energy-monitor.ts
Delta456: It should silently be ignored but may cause issues. I will check and update.

src/services/ios/dvt/instruments/energy-monitor.ts
Delta456: No, it shouldn't. It is synchronous. It is also rate limited by the device.

src/services/ios/dvt/instruments/energy-monitor.ts
```diff
@@ -0,0 +1,168 @@
...
+
+  /**
+   * Take a single energy snapshot for the given PIDs.
+   *
+   * {@link startSampling} must be called first — without it the device returns
+   * empty metrics. The call blocks until the device replies, so calling this
+   * in a tight loop does not spin-loop the CPU.
+   */
+  async sample(pids: number[]): Promise<EnergyMonitorSample> {
+    return await this.sampleOnce(pids);
```
mykola-mokhnach, line 104: would it make sense to throw here if `this.sampling` is false?

mykola-mokhnach (approved):

src/services/ios/dvt/instruments/energy-monitor.ts
Delta456: Yes, it would make sense. Will add

Delta456: added

harsha509 (approved):
LGTM!

PR #256, opened 2026-06-27
feat: Add `com.apple.coredevice.pasteboardservice`

mykola-mokhnach:
## Summary
- Add `PasteboardService` for CoreDevice pasteboard access with a focused public API: `getText`/`setText`, `getUrl`/`setUrl`, and `getImage`/`setImage`.
- Add raw CoreDevice request/response support for direct XPC services that do not use the standard CoreDevice invocation envelope.
- Wire pasteboard startup through `Services.startPasteboardService()` and public root exports while keeping raw pasteboard constants, UTIs, builders, and snapshot helpers internal.


src/services/ios/pasteboard/index.ts
Delta456, line 1: The file has too many constants. I would prefer moving them to a separate file like `const.ts`

mykola-mokhnach: extracted constants and types

mykola-mokhnach:
weirdly format checks pass on my local

Delta456:
> weirdly format checks pass on my local

Should be fixed in https://github.com/appium/appium-ios-remotexpc/pull/257

src/services/ios/pasteboard/index.ts
```diff
@@ -0,0 +1,234 @@
...
+
+  /**
+   * Pull the pasteboard and return the first decodable URL string.
+   */
+  async getUrl(
+    pasteboardName = GENERAL_PASTEBOARD,
+  ): Promise<string | undefined> {
+    return PasteboardService.extractString(
+      await this.get(pasteboardName),
+      URL_UTIS,
```
harsha509, line 57: so are we returning, plain text as url ? because i see `getUrl `extracts over `URL_UTIS`, which includes the three text UTIs, so it matches plain-text items that are not URLs.

src/services/ios/pasteboard/index.ts
```diff
@@ -0,0 +1,234 @@
...
+      actionIdentifier: PASTEBOARD_COMMAND.SET,
+    });
+  }
+}
+
+function pickSnapshot(
+  snapshotOrReply: XPCDictionary | PasteboardSnapshot | PasteboardPullReply,
+): PasteboardSnapshot {
+  const pasteboard = asDictionary(snapshotOrReply.pasteboard);
+  return (pasteboard ?? snapshotOrReply) as PasteboardSnapshot;
```
harsha509, line 217: When `pasteboard `is absent, the entire reply is cast to `PasteboardSnapshot `and .items is read off it.

If the real reply nests the snapshot under a different key, `snapshot.items` is `undefined`, `extract* `sees an empty list, and every getter returns `undefined` with no error.

src/services/ios/pasteboard/index.ts
mykola-mokhnach: now it returns URL instance

src/services/ios/pasteboard/index.ts
mykola-mokhnach: I think this is ok to assume the pasteboard is empty if no items are there

harsha509 (approved):
Thank you @mykola-mokhnach !

PR #260, opened 2026-06-30
chore: Replace prettier with oxfmt

mykola-mokhnach:
oxfmt does the formatting practically instantly on all project files, while prettier takes several seconds to do the same

Later we also would like to migrate linting to use oxlint because of the same reason

Files that were actually modified in this PR:
- package.json
- .oxfmtrc.json

Main diff between the previous formatting rules: increased maximum line length to 120 chars and tuned sorting of imports


KazuCocoa (approved):

harsha509 (approved):
Thank you @mykola-mokhnach !

package.json
```diff
@@ -94,37 +110,20 @@
     "esmock": "^2.7.3",
     "husky": "^9.1.7",
     "lint-staged": "^17.0.2",
-    "prettier": "^3.9.1",
+    "oxfmt": "^0.x",
```
eglitise, line 113: Note: oxfmt is still in pre-release, with roughly weekly releases, so this syntax will pick up any breaking changes and would cause CI and staging failures. Still, from my experience, breaking changes are fairly rare.

navin772 (approved):

PR #267, opened 2026-07-03
feat(coredevice): add coredevice device info service

navin772:
Functions added:

- `startCoreDeviceInfoService(udid)` → `CoreDeviceInfoService`
- `getDeviceInfo()` — OS/build/hardware identity, CPU cores, boot state, etc.
- `getDisplayInfo()` — display geometry: native size, scale, refresh rate, color gamut, orientation (useful for coordinate mapping / screenshots)
- `getLockState()` — device lock/passcode state
- `queryMobileGestalt(keys)` — query arbitrary MobileGestalt keys

Limitations, while testing I got the following results in iOS real device, same result using pymobiledevice3, it might be gating on iOS device end:
- `getLockState` ❌ returns `CoreDevice.ActionError 2` ("is not implemented") on this iOS version.
- `queryMobileGestalt` ❌ returns `CoreDevice.ActionError 5` ("This operation cannot be performed") — gated to privileged/personalized contexts.


src/services/ios/device-info/index.ts
```diff
@@ -0,0 +1,98 @@
...
+   * ⚠️ May be gated: Apple restricts gestalt queries to specific (personalized)
+   * device contexts, so on a normal device this can return an error or empty
+   * values. Pass `timeoutMs` via `options` to bound the wait.
+   */
+  async queryMobileGestalt(keys: string[], options: CoreDeviceInvokeOptions = {}): Promise<XPCDictionary> {
+    return asDict(await this.invoke(FEATURE_QUERY_MOBILEGESTALT, {keys}, options));
+  }
+}
+
+function asDict(value: XPCValue): XPCDictionary {
```
mykola-mokhnach, line 91: I have already seen this helper somewhere else. Can we avoid unnecessary duplication?

mykola-mokhnach (approved):

src/services/ios/device-info/index.ts
navin772: Will refactor in a separate PR

PR #268, opened 2026-07-03
refactor(xpc): centralize `asDictionary` helper

navin772:
as discussed in - https://github.com/appium/appium-ios-remotexpc/pull/267#discussion_r3520226534

### DIFF
--- a/src/lib/remote-xpc/xpc-value.ts
+++ b/src/lib/remote-xpc/xpc-value.ts
@@ -1,0 +1,9 @@
+import type {XPCDictionary, XPCValue} from '../types.js';
+
+/**
+ * Coerces an XPC value to a dictionary, or `undefined` when it is not a plain
+ * object (an array, primitive, `null`, or `undefined`).
+ */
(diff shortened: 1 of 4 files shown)


src/lib/remote-xpc/xpc-value.ts
```diff
@@ -0,0 +1,9 @@
+import type {XPCDictionary, XPCValue} from '../types.js';
+
+/**
+ * Coerces an XPC value to a dictionary, or `undefined` when it is not a plain
+ * object (an array, primitive, `null`, or `undefined`).
+ */
+export function asDictionary(value: XPCValue | undefined): XPCDictionary | undefined {
+  return value && typeof value === 'object' && !Array.isArray(value) ? (value as XPCDictionary) : undefined;
```
mykola-mokhnach: you could also use util.isPlainObject helper from appium support package

mykola-mokhnach (approved):

src/lib/remote-xpc/xpc-value.ts
navin772: updated, nice catch!

PR #269, opened 2026-07-03
feat(coredevice): add device control service to control orientation

navin772:
Adds ability to rotate device orientation.

Only possible to rotate when iOS orientation lock is not set.

### DIFF
--- a/src/services/ios/device-control/index.ts
+++ b/src/services/ios/device-control/index.ts
@@ -15,7 +15,13 @@ export interface DeviceOrientationState {
   currentDeviceOrientation?: string;
   /** The most recent non-flat orientation (ignores face-up/face-down). */
   currentDeviceNonFlatOrientation?: string;
-  /** Whether iOS orientation lock is engaged; when `true`, rotation may be prevented. */
+  /**
+   * Device-reported orientation-lock flag. Do **not** rely on this to detect
+   * Control Center Rotation Lock: on iOS 26.5 it was observed to stay `false`
+   * even while that lock was active and preventing rotation. To tell whether a
+   * rotation was actually applied, compare {@link currentDeviceOrientation}
+   * before and after {@link DeviceControlService.rotate}.
+   */
   currentDeviceOrientationLocked?: boolean;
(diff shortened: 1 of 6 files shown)


navin772:
Hi @mykola-mokhnach, could you review this PR when you get time?

src/services/ios/device-control/index.ts
```diff
@@ -0,0 +1,69 @@
...
+export interface DeviceOrientationState {
+  /**
+   * The resulting device orientation, e.g. `'portrait'`, `'landscapeLeft'`,
+   * `'portraitUpsideDown'`, `'landscapeRight'`.
+   */
+  currentDeviceOrientation?: string;
+  /** The most recent non-flat orientation (ignores face-up/face-down). */
+  currentDeviceNonFlatOrientation?: string;
+  /** Whether iOS orientation lock is engaged; when `true`, rotation may be prevented. */
+  currentDeviceOrientationLocked?: boolean;
```
mykola-mokhnach, line 25: is there a way to know if device orientation is locked in advance or we can only know about it after getting an exception by calling `rotate`?

test/unit/device-control/device-control-service.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+  });
+
+  it('closes the active transport', async function () {
+    const fake = new FakeTransport(() => ({
+      currentDeviceOrientation: 'portrait',
+    }));
+    const service = new TestDeviceControlService(fake);
+
+    await service.rotate('left');
+    await service.close();
```
mykola-mokhnach: the service should be closed inside a finally block, otherwise it may never happen if the test fails

src/services/ios/device-control/index.ts
navin772: No, we can't query the orientation lock, also it does not throws, just silently prevents it. We can only compare `currentDeviceOrientation` before/after a rotate. Will document this behavior

mykola-mokhnach (approved):

PR #270, opened 2026-07-06
feat(dvt): add activity trace tap service

Delta456:
This PR adds DVT Instrument Activity Trace Tap Service. Implementation inspired from https://github.com/doronz88/pymobiledevice3/blob/master/pymobiledevice3/services/dvt/instruments/activity_trace_tap.py


src/services/ios/dvt/instruments/activity-trace-tap.ts
```diff
@@ -0,0 +1,527 @@
...
+  private receiveAbortController: AbortController | null = null;
+
+  constructor(
+    dvt: DVTSecureSocketProxyService,
+    private readonly options: ActivityTraceTapOptions = {},
+  ) {
+    super(dvt);
+  }
+
+  async start(): Promise<void> {
```
mykola-mokhnach, line 85: not sure why to make start/stop public if they are anyway invoked in the `messages` generator.

Perhaps, only the public `stop` makes sense there to stop the async loop

src/services/ios/dvt/instruments/activity-trace-tap.ts
```diff
@@ -0,0 +1,527 @@
...
+   * Async generator that yields decoded log entries as they arrive.
+   * Starts the tap if not already started; stops when the generator is
+   * returned or {@link stop} is called.
+   */
+  async *messages(): AsyncGenerator<ActivityTraceMessage, void, unknown> {
+    await this.start();
+    // Anchors mach continuous time (ns) to wall clock on the first message.
+    let bootEpochMs: number | null = null;
+    try {
+      while (!this.stopRequested) {
```
mykola-mokhnach, line 144: is this loop also stopped properly after dvt service is closed?

src/services/ios/dvt/instruments/activity-trace-tap.ts
```diff
@@ -0,0 +1,527 @@
...
+          }
+          yield msg;
+        }
+      }
+    } finally {
+      await this.stop();
+    }
+  }
+
+  private *parseFrame(frame: Buffer): Generator<ActivityTraceMessage> {
```
mykola-mokhnach, line 192: can we simplify this long helper?

src/services/ios/dvt/instruments/activity-trace-tap.ts
```diff
@@ -0,0 +1,527 @@
...
+          columns.push(decodeStr(item as Buffer));
+        } else {
+          columns.push('');
+        }
+      }
+    }
+    this.tables.set(tableId, { name, columns });
+  }
+
+  private parseEndRow(word: number): ActivityTraceMessage | null {
```
mykola-mokhnach, line 343: same here. The method seems to be very long. 

I would also expect this complex logic to be unit tested properly

src/services/ios/dvt/instruments/activity-trace-tap.ts
```diff
@@ -0,0 +1,527 @@
...
+  }
+}
+
+function decodeStr(buf: Buffer): string {
+  const nullIdx = buf.indexOf(0);
+  return (nullIdx >= 0 ? buf.subarray(0, nullIdx) : buf).toString('utf8');
+}
+
+function decodeMessageFormat(message: StackItem[]): string {
+  let s = '';
```
mykola-mokhnach: consider pushing to array and joining it at the end rather than catenating a string (each string catenation creates a new object).

src/services/ios/dvt/index.ts
```diff
@@ -266,7 +271,7 @@ export class DVTSecureSocketProxyService extends BaseService {
   private async performHandshake(): Promise<void> {
     const args = new MessageAux();
     args.appendObj({
-      'com.apple.private.DTXBlockCompression': 2,
+      'com.apple.private.DTXBlockCompression': 0,
```
mykola-mokhnach, line 274: may this change break other services?

test/integration/dvt_instruments/activity-trace-tap-test.ts
```diff
@@ -0,0 +1,328 @@
...
+
+const KNOWN_MESSAGE_TYPES = new Set([
+  'Default',
+  'Info',
+  'Debug',
+  'Error',
+  'Fault',
+]);
+
+function requireUdid(ctx: Mocha.Context): void {
```
mykola-mokhnach: we do not use mocha anymore. Use `node:test` instead.

Also there is the shared `requireDeviceUdid` method

Delta456: Yes the file is now deleted.

src/services/ios/dvt/index.ts
Delta456: I accidentally changed to 0 in my previous PRs which broke few services so I had to revert this.

src/services/ios/dvt/instruments/activity-trace-tap.ts
Delta456: I will see what can be done

src/services/ios/dvt/instruments/activity-trace-tap.ts
Delta456: We do this for other message services. This can be breaking if only one service does not follow the pattern.

src/services/ios/dvt/instruments/activity-trace-tap.ts
Delta456: Yes, it should be. I will cross check though

src/services/ios/dvt/instruments/activity-trace-tap.ts
Delta456: Sounds like a good idea

src/services/ios/dvt/instruments/activity-trace-tap.ts
Delta456: Done

src/services/ios/dvt/instruments/activity-trace-tap.ts
Delta456: Done

src/services/ios/dvt/instruments/activity-trace-tap.ts
Delta456: Done

src/services/ios/dvt/instruments/activity-trace-tap.ts
Delta456: Yes the loop is stopped properly.

src/services/ios/dvt/index.ts
mykola-mokhnach: I'm asking because of https://github.com/appium/appium-ios-remotexpc/pull/254#discussion_r3481020183

mykola-mokhnach (approved):

src/services/ios/dvt/index.ts
Delta456: It was a miss on my part. Energy monitor works with this too.

test/integration/dvt-instruments/activity-trace-tap.spec.ts
```diff
@@ -0,0 +1,246 @@
...
+
+        const terminal = await Promise.race([
+          (async () => {
+            let result = await nextPromise;
+            while (!result.done) {
+              result = await iterator.next();
+            }
+            return result;
+          })(),
+          new Promise<never>((_, reject) =>
```
navin772: Could you fix lint warning here and on line 214:

```
Promise constructor parameters must be named to match "^_?resolve$"
```

Delta456: done

PR #271, opened 2026-07-07
feat: Add codegraph

mykola-mokhnach:
https://github.com/colbymchenry/codegraph#why-codegraph


navin772 (approved):
Tested it out, looks good enough.

One caveat is that it injects sources even on non-code questions which might consume more tokens, the LLM disregards these sources so shouldn't be of much issue.

PR #273, opened 2026-07-07
fix(notification-proxy): make observe/post fire-and-forget

Delta456:
## Summary
- `ObserveNotification`/`PostNotification` have no per-command acknowledgement on the real wire protocol (matching pymobiledevice3's own `send_plist`-only implementation), but `observe()`/`post()` waited for a reply via `sendPlistRequest()`.
- The RSD shim's initial `StartService` greeting was never drained, so the first call on a fresh connection would accidentally consume that greeting as if it were the reply, while every call after that hung for the full timeout and threw.
- Fix: drain the greeting once when the connection is opened, and send `ObserveNotification`/`PostNotification` without waiting for a reply — matching the real device protocol and pymobiledevice3's reference implementation.

Confirmed live against a real device: previously, the first `observe()`/`post()` call on a connection "succeeded" in ~60ms by returning the unrelated greeting object, and every subsequent call hung for the full 10s timeout before throwing.

Not a breaking change in practice: the `Promise<PlistDictionary>` signature is unchanged (`{}` is a valid value), and the old return content was never a working/documented contract. Checked the real downstream consumer (`appium-xcuitest-driver`) and confirmed it never reads the return value of either call.

## Test plan
- [x] Unit/integration test `test/integration/notification-proxy.spec.ts` updated to stop asserting on `post()`'s return value, pulling the relayed notification from `expectNotifications()` instead
- [x] Verified live against a real device (multiple `observe`/`post` calls in sequence no longer hang)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/services/ios/notification-proxy/index.ts
+++ b/src/services/ios/notification-proxy/index.ts
@@ -38,23 +38,29 @@ class NotificationProxyService extends BaseService implements NotificationProxyS
 
   /**
    * Observe a notification
+   *
+   * `ObserveNotification` has no per-command acknowledgement on the wire (matching
+   * notification_proxy's real protocol), so this only sends the request; it does not
+   * wait for or return a device reply.
    * @param notification The notification name to subscribe to
-   * @returns Promise that resolves when the subscription request is sent
+   * @returns Promise that resolves once the subscription request has been sent
    */
   async observe(notification: string): Promise<PlistDictionary> {
-    if (!this._conn) {
-      this._conn = await this.connectToNotificationProxyService();
-    }
-    const request = this.createObserveNotificationRequest(notification);
-    const result = await this.sendPlistDictionary(request);
+    const conn = await this.connectToNotificationProxyService();
+    conn.sendPlist(this.createObserveNotificationRequest(notification));
     this._pendingNotificationsObservationSet.add(notification);
-    return result;
+    return {};
   }
 
   /**
    * Post a notification
+   *
+   * Like `ObserveNotification`, `PostNotification` has no per-command acknowledgement
+   * on the wire, so this only sends the request; it does not wait for a device reply.
+   * Waiting here would also race any concurrent `expectNotifications()`/`expectNotification()`
+   * reader on the same connection for whatever message arrives next.
    * @param notification The notification name to post
-   * @returns Promise that resolves when the post request is sent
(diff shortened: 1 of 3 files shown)


src/services/ios/notification-proxy/index.ts
```diff
@@ -38,23 +38,29 @@ class NotificationProxyService extends BaseService implements NotificationProxyS
...
-    if (!this._conn) {
-      this._conn = await this.connectToNotificationProxyService();
-    }
-    const request = this.createObserveNotificationRequest(notification);
-    const result = await this.sendPlistDictionary(request);
+    const conn = await this.connectToNotificationProxyService();
+    conn.sendPlist(this.createObserveNotificationRequest(notification));
     this._pendingNotificationsObservationSet.add(notification);
-    return result;
+    return {};
```
mykola-mokhnach, line 53: should we change the return type to `type EmptyObject = Record<string, never>;` ?

src/services/ios/notification-proxy/index.ts
```diff
@@ -64,11 +70,10 @@ class NotificationProxyService extends BaseService implements NotificationProxyS
...
       throw new Error('You must call observe() before posting notifications.');
     }
-    this._conn = await this.connectToNotificationProxyService();
-    const request = this.createPostNotificationRequest(notification);
-    const result = await this.sendPlistDictionary(request);
+    const conn = await this.connectToNotificationProxyService();
+    conn.sendPlist(this.createPostNotificationRequest(notification));
     this._pendingNotificationsObservationSet.delete(notification);
-    return result;
+    return {};
```
mykola-mokhnach, line 77: same as above

mykola-mokhnach (approved):

src/services/ios/notification-proxy/index.ts
Delta456: Yes we can

Delta456: Done

src/services/ios/notification-proxy/index.ts
Delta456: Done

navin772 (approved):

mykola-mokhnach (approved):

PR #274, opened 2026-07-08
feat(coredevice): add `getOrientation` method

navin772:
Workaround to - https://github.com/appium/appium-ios-remotexpc/pull/269#discussion_r3538088984 

Functionality:

Adds a method `getOrientation()`, it rotates the device one 90° step and immediately reverts it:
  - If the orientation changed, rotation is **unlocked**.
  - If it stayed the same, rotation is **locked**.

It is kind of a helper for convenient use.

### DIFF
--- a/src/services/ios/device-control/index.ts
+++ b/src/services/ios/device-control/index.ts
@@ -27,6 +27,27 @@ export interface DeviceOrientationState {
 }
 
 /**
+ * Result of {@link DeviceControlService.getOrientation}: the device's current
+ * orientation together with an empirically-determined rotation-lock flag.
+ */
+export interface DeviceOrientationInfo {
+  /**
+   * Current device orientation, e.g. `'portrait'`, `'landscapeLeft'`. Because
+   * the probe restores the device, this equals the orientation the device had
+   * before {@link DeviceControlService.getOrientation} was called.
+   */
+  orientation?: string;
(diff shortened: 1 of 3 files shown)


src/services/ios/device-control/index.ts
```diff
@@ -77,6 +98,37 @@ export class DeviceControlService extends CoreDeviceService {
...
+   * The two rotations cancel out, so the device ends where it began — though it
+   * does briefly rotate and rotate back while the probe runs.
+   *
+   * @example
+   * ```ts
+   * const {orientation, locked} = await deviceControl.getOrientation();
+   * console.log(`device is ${orientation}${locked ? ' (rotation locked)' : ''}`);
+   * ```
+   */
+  async getOrientation(): Promise<DeviceOrientationInfo> {
```
mykola-mokhnach, line 123: this method has an unexpected side effect. I'd prefer to not add it to the public interface

navin772: is the rotations made to determine the current orientation the side effect? should we keep it as it is documenting the behaviour

mykola-mokhnach: exactly. Performing unexpected rotation may unexpectedly change the state of the application under test

navin772: Understandable, closing this PR.

PR #277, opened 2026-07-14
feat(coredevice): add configuration service (appearance & accessibility)

navin772:
Adds `ConfigurationService` (`com.apple.coredevice.configuration`) - the `devicectl` backend for the appearance and accessibility knobs Xcode toggles.

## API
| Setting | Methods |
|---|---|
| Appearance (dark/light) | `getUserInterfaceStyle` / `setUserInterfaceStyle` |
| Dynamic Type size | `getDeviceTextSize` / `setDeviceTextSize` |
| Reduce Motion | `getReduceMotion` / `setReduceMotion` |
| Reduce Transparency | `getReduceTransparency` / `setReduceTransparency` |
| Increase Contrast | `setIncreaseContrast` (no getter) |
| Color filter | `getColorFilter` / `setColorFilter` |
| Layout borders | `getShowBorders` / `setShowBorders` |
| Liquid glass (iOS 26) | `setLiquidGlassOpacity` |

## Device-gated behaviors (surfaces a descriptive `CoreDeviceError`)
- **Liquid glass**: hardware-gated - unsupported devices reject with error 21035 even on iOS 26 iPhone 11, this might work in newer models.
- **Color filter `intensity`**: rejected with error 21056 on some devices; omit it. 
  `setColorFilter(true, {filterType: 'Grayscale'})` works.

### DIFF
--- a/src/services/ios/configuration/index.ts
+++ b/src/services/ios/configuration/index.ts
@@ -1,6 +1,6 @@
 import {asDictionary} from '../../../lib/remote-xpc/xpc-value.js';
 import type {XPCDictionary} from '../../../lib/types.js';
-import {type CoreDeviceInvokeOptions, CoreDeviceService} from '../core-device/core-device-service.js';
+import {type CoreDeviceInvokeOptions, CoreDeviceError, CoreDeviceService} from '../core-device/core-device-service.js';
 
 const ACTION = {
   GET_USER_INTERFACE_STYLE: 'com.apple.coredevice.action.getuserinterfacestyle',
@@ -12,6 +12,7 @@ const ACTION = {
   SET_DEVICE_TEXT_SIZE: 'com.apple.coredevice.action.setdevicetextsize',
   GET_REDUCE_MOTION: 'com.apple.coredevice.action.getreducemotion',
   SET_REDUCE_MOTION: 'com.apple.coredevice.action.setreducemotion',
+  GET_INCREASE_CONTRAST: 'com.apple.coredevice.action.getdeviceincreasecontrast',
   SET_INCREASE_CONTRAST: 'com.apple.coredevice.action.setdeviceincreasecontrast',
   GET_SHOW_BORDERS: 'com.apple.coredevice.action.getshowborders',
   SET_SHOW_BORDERS: 'com.apple.coredevice.action.setshowborders',
@@ -21,6 +22,31 @@ const ACTION = {
 
 /** Device appearance: `'dark'` or `'light'`. */
 export type UserInterfaceStyle = 'dark' | 'light';
+
+/**
+ * Dynamic Type content-size names accepted by the daemon's `setdevicetextsize`
+ * action. These are the seven standard sizes.
+ */
+const DEVICE_TEXT_SIZES = [
+  'extraSmall',
+  'small',
+  'medium',
+  'large',
+  'extraLarge',
+  'extraExtraLarge',
+  'extraExtraExtraLarge',
+] as const;
+
+/** A valid Dynamic Type content-size name. See {@link DEVICE_TEXT_SIZES}. */
+export type DeviceTextSize = (typeof DEVICE_TEXT_SIZES)[number];
+
+/**
+ * Color-filter presets that enable from `filterType` alone.
+ */
+const COLOR_FILTER_TYPES = ['Grayscale', 'Protanopia', 'Deuteranopia', 'Tritanopia'] as const;
+
+/** A color-filter preset that enables from a name alone. See {@link COLOR_FILTER_TYPES}. */
+export type ColorFilterType = (typeof COLOR_FILTER_TYPES)[number];
 
 /** Color-filter state returned by {@link ConfigurationService.getColorFilter}. */
 export interface ColorFilterState {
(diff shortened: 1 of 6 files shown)


src/services/ios/configuration/index.ts
```diff
@@ -0,0 +1,212 @@
...
+ */
+export class ConfigurationService extends CoreDeviceService {
+  static readonly RSD_SERVICE_NAME = 'com.apple.coredevice.configuration';
+
+  constructor(udid: string) {
+    super(udid, ConfigurationService.RSD_SERVICE_NAME);
+  }
+
+  /** Returns the active appearance — `'dark'` or `'light'`. */
+  async getUserInterfaceStyle(options: CoreDeviceInvokeOptions = {}): Promise<UserInterfaceStyle> {
```
harsha509, line 111: if style is undefined, are we setting tro dark by default?

src/services/ios/configuration/index.ts
```diff
@@ -0,0 +1,212 @@
...
+
+  /** Returns the Dynamic Type size name (e.g. `'medium'`, `'large'`). */
+  async getDeviceTextSize(options: CoreDeviceInvokeOptions = {}): Promise<string | undefined> {
+    const output = await this.action(ACTION.GET_DEVICE_TEXT_SIZE, {}, options);
+    const size = asDictionary(asDictionary(output.textSize)?.size);
+    return size ? Object.keys(size)[0] : undefined;
+  }
+
+  /** Sets the Dynamic Type size by name (`'medium'`, `'large'`, …). */
+  async setDeviceTextSize(size: string, options: CoreDeviceInvokeOptions = {}): Promise<void> {
```
harsha509: looks like it accepts any string unvalidated

src/services/ios/configuration/index.ts
```diff
@@ -0,0 +1,212 @@
...
+   * rejected with `com.apple.dt.CoreDeviceError 21056` ("The color filter
+   * intensity value is not valid."), surfaced here as a {@link CoreDeviceError}.
+   * Omit it unless you know the target device accepts it. Some filter presets
+   * (e.g. `'colorTint'`) also need extra parameters and will not enable from
+   * `filterType` alone.
+   */
+  async setColorFilter(enabled: boolean, options: SetColorFilterOptions & CoreDeviceInvokeOptions = {}): Promise<void> {
+    const {filterType, intensity, ...invokeOptions} = options;
+    const colorFilter: XPCDictionary = {enabled};
+    if (enabled) {
```
harsha509, line 176: if false, we are dropping off options here ?
just wanted to check if this is intentional

src/services/ios/configuration/index.ts
```diff
@@ -0,0 +1,212 @@
...
+    return asDictionary(output[key])?.enabled === true;
+  }
+}
+
+/**
+ * Rounds `value` through IEEE-754 binary32.
+ * The daemon's Swift decoders reject float knobs whose low mantissa bits do not
+ * fit in Float32; quantizing here produces a bit pattern the device accepts.
+ */
+function toFloat32(value: number): number {
```
harsha509: there are only 2 places it is used, it could sit directly on  return `Math.fround(value);`

src/services/ios/configuration/index.ts
navin772: will add a supported list and validate against it

src/services/ios/configuration/index.ts
navin772: the device only returns either light or dark, will gate to these 2 values only

src/services/ios/configuration/index.ts
navin772: yes, its intentional

src/services/ios/configuration/index.ts
```diff
@@ -0,0 +1,256 @@
...
+  constructor(udid: string) {
+    super(udid, ConfigurationService.RSD_SERVICE_NAME);
+  }
+
+  /** Returns the active appearance — `'dark'` or `'light'`. */
+  async getUserInterfaceStyle(options: CoreDeviceInvokeOptions = {}): Promise<UserInterfaceStyle> {
+    const output = await this.action(ACTION.GET_USER_INTERFACE_STYLE, {}, options);
+    const style = output.style;
+    if (style !== 'dark' && style !== 'light') {
+      throw new CoreDeviceError(`Unexpected user-interface style from device: ${String(style)}`, output);
```
mykola-mokhnach: this is not very flexible for the future changes

mykola-mokhnach: same below

src/services/ios/configuration/index.ts
```diff
@@ -0,0 +1,256 @@
...
+    if (enabled) {
+      if (filterType === undefined) {
+        throw new TypeError('filterType is required when enabling the color filter');
+      }
+      if (!isOneOf(COLOR_FILTER_TYPES, filterType)) {
+        throw new TypeError(`filterType must be one of ${COLOR_FILTER_TYPES.join(', ')}, got '${String(filterType)}'`);
+      }
+      colorFilter.filterType = {name: filterType};
+      if (intensity !== undefined) {
+        colorFilter.intensity = Math.fround(intensity);
```
mykola-mokhnach, line 185: since the `fround` trick is repeated more than once, we could create a common helper for it, whose function name would be self-documenting

mykola-mokhnach (approved):

src/services/ios/configuration/index.ts
navin772: There was a suggestion to inline it since used only at 2 sites - https://github.com/appium/appium-ios-remotexpc/pull/277#discussion_r3580008828

src/services/ios/configuration/index.ts
navin772: Updated to `export type UserInterfaceStyle = 'dark' | 'light' | (string & {});`, it now allows any string

PR #283, opened 2026-07-30
feat(display): add CoreDevice DisplayService for HEVC/AAC screen and audio capture

navin772:
Adds `com.apple.coredevice.displayservice` — the backend behind Xcode's device mirroring. Captures the device's screen as HEVC and its system audio as AAC-ELD, both streamed over RTP across the tunnel.

**Requires iOS 27.0+ and a mounted DDI.** Below iOS 27 the daemon answers the capability queries but rejects every stream with `CoreDeviceError` 9021 ("Remote control requires iOS 27.0 or later"). Use `isStreamingSupported()` as a pre-flight check.

## Usage

```ts
import {Services, recordScreenAndAudioToFiles} from 'appium-ios-remotexpc';

const display = await Services.startDisplayService(udid);
try {
  if (!(await display.isStreamingSupported())) return;

  // Screen + audio, written as two tracks
  const result = await recordScreenAndAudioToFiles(display, {
    videoPath: '/tmp/screen.h265',
    audioPath: '/tmp/screen.m4a',
    durationMs: 45_000,
  });
  console.log(result.ffmpegCommand); // ready-to-run mux, frame rate filled in
} finally {
  await display.close();
}
```

Video or audio alone:

```ts
await recordScreenToFile(display, '/tmp/screen.h265', {durationMs: 45_000});
await recordAudioToFile(display, '/tmp/audio.m4a', {durationMs: 45_000});
```

Live frames instead of a file:

```ts
const capture = await ScreenStreamCapture.start(display, {displayId: 1});
try {
  for await (const au of capture.accessUnits()) {
    // au.nals, au.isKeyFrame, au.timestamp
  }
} finally {
  await capture.stop();
}
```

## Notes for users

- **Muxing is left to the caller.** `recordScreenAndAudioToFiles` returns `ffmpegCommand` with the measured frame rate substituted. Annex-B carries no timestamps, so `-fflags +genpts` and an explicit `-r` are both required, and `-c copy` avoids decoding AAC-ELD entirely.


src/services/ios/display/audio/aac-eld.ts
```diff
@@ -0,0 +1,91 @@
...
+  audioSpecificConfig: Buffer;
+}
+
+/** The format the device streams. */
+export const AAC_ELD_FORMAT: AacEldFormat = {
+  sampleRate: AAC_ELD_SAMPLE_RATE,
+  channels: AAC_ELD_CHANNELS,
+  framesPerPacket: AAC_ELD_FRAMES_PER_PACKET,
+  audioSpecificConfig: AAC_ELD_ASC_48K_STEREO_480,
+};
```
mykola-mokhnach, line 88: as const?

src/services/ios/display/access-unit-assembler.ts
```diff
@@ -0,0 +1,187 @@
...
+    this.statsInternal.packetsReceived += 1;
+
+    if (this.lastSequence !== undefined && !isNextSequence(this.lastSequence, packet.sequence)) {
+      // A gap means a fragment is missing: any in-flight NAL can no longer be
+      // completed, and the picture being assembled is incomplete.
+      const missing = (packet.sequence - this.lastSequence - 1) & 0xffff;
+      this.statsInternal.packetsLost += missing;
+      this.depacketizer.reset();
+      this.currentCorrupt = true;
+      log.debug(`RTP gap: ${missing} packet(s) lost before sequence ${packet.sequence}`);
```
mykola-mokhnach: this might be logged to frequently

src/services/ios/display/audio-stream-capture.ts
```diff
@@ -0,0 +1,262 @@
...
+        const packet = parseRtpPacket(datagram);
+        if (!packet || packet.payload.length === 0) {
+          continue; // RTCP, malformed, or an empty keep-alive.
+        }
+        this.statsInternal.packetsReceived += 1;
+
+        if (this.lastSequence !== undefined && !isNextSequence(this.lastSequence, packet.sequence)) {
+          const missing = (packet.sequence - this.lastSequence - 1) & 0xffff;
+          this.statsInternal.packetsLost += missing;
+          log.debug(`Audio RTP gap: ${missing} packet(s) lost before sequence ${packet.sequence}`);
```
mykola-mokhnach: same as above, this might be logged too often

src/services/ios/display/audio-stream-capture.ts
```diff
@@ -0,0 +1,262 @@
...
+ */
+export async function recordAudioToFile(
+  service: DisplayService,
+  outputPath: string,
+  options: RecordAudioOptions = {},
+): Promise<RecordAudioResult> {
+  const {durationMs = 5000, ...captureOptions} = options;
+
+  const capture = await AudioStreamCapture.start(service, captureOptions);
+  const accessUnits: Buffer[] = [];
```
mykola-mokhnach: can it go out of memory?

src/services/ios/display/audio-stream-capture.ts
```diff
@@ -0,0 +1,262 @@
...
+    }
+  } finally {
+    clearTimeout(timer);
+    await capture.stop().catch((error: unknown) => {
+      log.debug(`Failed to stop the audio stream cleanly: ${error instanceof Error ? error.message : String(error)}`);
+    });
+  }
+
+  const m4a = buildM4a(accessUnits);
+  await writeFile(outputPath, m4a);
```
mykola-mokhnach: not sure if we can really do this. By keeping these packets for too long in memory we can quickly go OOM. Can we do streaming instead?

src/services/ios/display/av-capture.ts
```diff
@@ -0,0 +1,238 @@
...
+  let audioCapture: AudioStreamCapture;
+  try {
+    audioCapture = await AudioStreamCapture.start(service, {clientSessionId});
+  } catch (error) {
+    await videoCapture.stop().catch((): void => undefined);
+    throw error;
+  }
+
+  const videoOut = createWriteStream(videoPath);
+  const audioUnits: Buffer[] = [];
```
mykola-mokhnach: same as above

src/services/ios/display/audio-stream-capture.ts
navin772: Streaming can be done but its a completely new functionality that I want to implement in a separate PR

src/services/ios/display/av-capture.ts
```diff
@@ -0,0 +1,238 @@
...
+ * Builds the `ffmpeg` command that combines the two tracks.
+ *
+ * `-fflags +genpts` with an explicit `-r` is required because the Annex-B video
+ * carries no timestamps — without them a muxer produces an empty audio track.
+ */
+function buildFfmpegCommand(options: {videoPath: string; audioPath: string; frameRate: number}): string {
+  const {videoPath, audioPath, frameRate} = options;
+  const output = videoPath.replace(/\.[^.]+$/, '') + '.mp4';
+  return (
+    `ffmpeg -y -fflags +genpts -r ${frameRate} -i '${videoPath}' -i '${audioPath}' ` +
```
mykola-mokhnach: prefer just returning array of arguments instead of a full string. This will be more flexible and portable as ffmpeg binary itself might be named anything and be located anywhere.

Also, it makes sense to isolate this logic in a separate class/module, se we could potentially add more beckends, like gstreamer

mykola-mokhnach:
I think it makes sense to group files in the `display` folder into smaller packages. like video, audio, rtc, etc.

src/services/ios/display/index.ts
```diff
@@ -0,0 +1,393 @@
...
+   *
+   * @param endpoint Where the device should send RTP, and its own address.
+   * @param options Display selection, session id and encoder knobs.
+   */
+  async startVideoStream(
+    endpoint: MediaStreamEndpoint,
+    options: StartVideoStreamOptions = {},
+  ): Promise<MediaStreamAnswer> {
+    const {
+      displayId = 1,
```
mykola-mokhnach, line 231: is there any API that allows to list displays info?

src/services/ios/display/index.ts
```diff
@@ -0,0 +1,393 @@
...
+    };
+
+    return this.startMediaStream(input, clientSessionId, {timeoutMs, actionIdentifier});
+  }
+
+  /**
+   * Stops **every** media stream currently running on the device and returns
+   * the stopped streams' ids (each is the stream's `RemoteSSRC`).
+   *
+   * The device offers no per-session stop: the request requires a `stopAll`
```
mykola-mokhnach, line 289: is it safe to run this API if no streaming is running or it is going to throw an exception?
does it also throw if streaming is not supported/ios version is too low?

src/services/ios/display/index.ts
```diff
@@ -0,0 +1,393 @@
...
+    endpoint: MediaStreamEndpoint,
+    negotiatorOffer: Buffer,
+    negotiationTimeoutSeconds: number,
+  ): XPCDictionary {
+    return {
+      clientSupportedFeatures: CLIENT_SUPPORTED_FEATURES,
+      direction: 'output',
+      negotiatorOffer,
+      receiverIP: endpoint.receiverIp,
+      receiverPort: BigInt(endpoint.receiverPort),
```
mykola-mokhnach, line 360: isn't port something between 1..0xFFFF? why do we need bigint for it?

src/services/ios/display/index.ts
mykola-mokhnach: same question to APIs above

src/services/ios/display/audio-stream-capture.ts
mykola-mokhnach: video already does that though

navin772: oh, I see what you meant, I was thinking about serving the stream live to a consumer, like a web serve.
I will see if its possible with audio to do streaming to disk.

src/services/ios/display/index.ts
navin772: no exception, just empty list, will update the method docstrings

src/services/ios/display/index.ts
navin772: Not in this service, but we can use `CoreDeviceInfoService.getDisplayInfo()`, will document this in docstring

src/services/ios/display/access-unit-assembler.ts
navin772: will fix this

src/services/ios/display/audio/aac-eld.ts
navin772: made them `readonly`

src/services/ios/display/audio-stream-capture.ts
navin772: updated audio to stream instead of write all to memory and then write to disk

src/services/ios/display/av-capture.ts
navin772: updated to stream now

src/services/ios/display/index.ts
navin772: It is intended, when we set `receiverPort` as number we get error:
`NSCocoaErrorDomain 4864: "Expected to decode UInt16 but found a OS_xpc_int64 instead"`

We are using BigInt for consistent wire-type so that we match the XPC the device expects.

src/services/ios/display/audio-stream-capture.ts
navin772: updated to stream to disk

Delta456 (approved):

harsha509 (approved):
LGTM!

src/services/ios/display/recording/mux-command.ts
```diff
@@ -0,0 +1,129 @@
...
+ * - `-fflags +genpts` with an explicit `-r` before the video input, because the
+ *   Annex-B stream has no timestamps — without them ffmpeg produces a file with
+ *   an empty audio track.
+ * - `-c copy`, so the AAC-ELD never has to be decoded. Most decoders get ELD
+ *   wrong (see `aac-eld.ts`), and copying sidesteps the question entirely.
+ * - `-tag:v hvc1` rather than the default `hev1`, which is what makes the
+ *   result playable in QuickTime and Safari.
+ */
+export const ffmpegMuxCommandBuilder: MuxCommandBuilder = {
+  name: 'ffmpeg',
```
mykola-mokhnach, line 79: ffmpeg name on Windows has .exe suffix

PR #286, opened 2026-08-05
fix: use an ephemeral relay port instead of a hardcoded 2222

mikebruin:
## Problem

`RelayService` binds a local TCP server on a hardcoded port (`2222`) purely so that it can
immediately connect to it over loopback. Because that port is a single global constant, any two
relays whose lifetimes overlap collide:

```
Error: listen EADDRINUSE: address already in use :::2222
```

Nothing outside the process ever references this port — `start()` binds it and `connect()`
immediately dials `127.0.0.1:<same port>` — so it never needed to be fixed.

This is easy to miss with a single device and unavoidable with several. Two code paths hit it
concurrently in normal use:

1. **`scripts/tunnel-creation.mjs` reconnect loops.** Startup processes devices sequentially, so no
   collision there. But `runReconnectAttempts` is per-UDID (`reconnectingByUdid` is keyed by UDID,
   not global), so N dropped devices retry *concurrently*, each once per second, each binding 2222.
   With `--reconnect-retries 0` they collide with each other indefinitely and never recover, even
   once the device is healthy again.

2. **`appium-xcuitest-driver`'s `withUsbMuxLockdown`**, which calls `createLockdownServiceByUDID`
   during session creation and takes the same relay path. Live test sessions therefore contend with
   each other and with the daemon above:

   ```
   org.openqa.selenium.WebDriverException: Failed to read lockdown via appium-ios-remotexpc
   USBMUX path for '<udid>': listen EADDRINUSE: address already in use :::2222
   ```

The knock-on effect is worse than the error itself. When a reconnect can never win the port, the
device stays absent from the tunnel registry. `getConnectedDevices()` in the XCUITest driver returns
tunnel-registry UDIDs only when the registry is reachable, so the driver then rejects a perfectly
healthy, USB-connected phone with `Unknown device or simulator UDID: '<udid>'` — an error that gives
no hint a port conflict is the cause.

## Evidence

From a 14-device iPhone farm running one shared `tunnel-creation.mjs` daemon since 2026-07-24. Two
devices stuck in unlimited reconnect, both on USB and visible to usbmuxd:

```
info DeviceManager Found device: DeviceID=21, SerialNumber=<udid>, ConnectionType=USB
ERR! TunnelCreation ❌ Failed to create tunnel for device <udid>: Error: listen EADDRINUSE: address already in use :::2222
WARN TunnelCreation Reconnecting dropped tunnel for <udid> (attempt 5807, unlimited mode)...
```

Attempt counters reached 7051 and 4029 for two devices. The log grew to 60 MB and the daemon
accumulated 242 minutes of CPU entirely from the retry/collide cycle. The tunnel registry served 12
entries while usbmuxd listed 14 devices — the two missing ones being exactly those stuck on the port.

## Reproduction

Needs one connected device and no tunnel:

```js
import {createLockdownServiceByUDID} from 'appium-ios-remotexpc';

const UDID = '<any connected udid>';
const one = async () => {
  try {
    const {lockdownService} = await createLockdownServiceByUDID(UDID);
    lockdownService.close();
    return 'ok';
  } catch (e) { return 'FAIL: ' + e.message; }
};
console.log(await Promise.all([one(), one()]));
```

Before: `[ 'FAIL: listen EADDRINUSE: address already in use :::2222', 'FAIL: ...' ]`
After: `[ 'ok', 'ok' ]`

## Change

Bind port `0` and read the OS-assigned port back off the listening server, so concurrent relays
cannot contend. Passing an explicit port still binds that port, so this is backward compatible —
only the default changes.

## Verification

Applied the equivalent change to the compiled output of 5.11.0 on the affected machine, with the
unpatched tunnel daemon still running and actively contending for 2222:

| scenario | before | after |
| --- | --- | --- |
| 2 concurrent `createLockdownServiceByUDID` | both `EADDRINUSE` | 2/2 succeeded |
| 6 concurrent | — | 6/6 succeeded |

On this branch: `npm run build`, `npm run lint`, and `npm run format:check` are clean. `npm run
test:unit` passes except for 15 pre-existing failures in `tunnel-registry-server.spec.ts`, which are
environmental on my machine — the spec writes the real strongbox `tunnelRegistryPort`, which is
root-owned there. Verified identical on unmodified `main` (15 failures, same `EACCES`), so they are
unrelated to this change.

## On test coverage

I did not add a unit test. `RelayService.start()` calls `createUsbmux()` intra-module before it
reaches the `listen()` call, so it cannot be intercepted with the existing `mockImport` helper, and
covering it would need either a real usbmuxd or an injectable usbmux factory. I did not want to
fold that refactor into a bug fix, but I'm happy to add it in whatever shape you prefer.

## Possibly worth separate issues

Neither is caused by this change, but both compound it:

1. **`tunnel-creation.mjs` never rescans for devices.** It calls `usbmux.listDevices()` once at
   startup then `await usbmux.close()`, with no attach/detach subscription. `devicesByUdid` is a
   startup snapshot and `reconnectTunnelByUdid` bails with `Cannot reconnect <udid>: device context
   not found` for anything absent from it, so a device plugged in after the daemon starts never
   receives a tunnel until a full restart.

2. **Unlimited reconnect has no backoff.** `runReconnectAttempts` retries on a flat `sleep(1000)`
   with no ceiling under `--reconnect-retries 0`. For a permanently-failing cause this is an
   unbounded hot loop — the 60 MB log and 242 CPU-minutes above. Exponential backoff, or throttling
   the log line, would make the failure mode much cheaper to live with.

### DIFF
--- a/src/lib/lockdown/index.ts
+++ b/src/lib/lockdown/index.ts
@@ -18,7 +18,8 @@ const deviceManagerLog = getLogger('DeviceManager');
 const LABEL = 'appium-internal';
 const DEFAULT_TIMEOUT = 5000;
 const DEFAULT_LOCKDOWN_PORT = 62078;
-const DEFAULT_RELAY_PORT = 2222;
+/** 0 lets the OS assign a free ephemeral port, so concurrent relays cannot collide. */
+const DEFAULT_RELAY_PORT = 0;
 /** RSD service name for lockdownd over a RemoteXPC tunnel (e.g. Apple TV Wi‑Fi). */
 const LOCKDOWN_REMOTE_UNTRUSTED = 'com.apple.mobile.lockdown.remote.untrusted';
 
--- a/src/lib/usbmux/index.ts
+++ b/src/lib/usbmux/index.ts
@@ -358,7 +358,7 @@ export class Usbmux extends BaseSocketService {
 export class RelayService {
   private readonly deviceID: string | number;
   private readonly devicePort: number;
-  private readonly relayPort: number;
+  private relayPort: number;
   private usbmuxClient: Socket | null;
   private server: Server | null;
 
@@ -366,9 +366,10 @@ export class RelayService {
    * Creates a new RelayService instance
    * @param deviceID - The device ID to connect to
    * @param devicePort - The port on the device to connect to
-   * @param relayPort - The local port to use for the relay server
+   * @param relayPort - The local port to use for the relay server. Defaults to 0, which lets the
+   * operating system assign a free ephemeral port so that concurrent relays cannot collide.
    */
-  constructor(deviceID: string | number, devicePort: number, relayPort: number = 2222) {
+  constructor(deviceID: string | number, devicePort: number, relayPort: number = 0) {
     this.deviceID = deviceID;
     this.devicePort = devicePort;
     this.relayPort = relayPort;
@@ -413,6 +414,12 @@ export class RelayService {
       }
 
       this.server.listen(this.relayPort, () => {
+        // Read back the port the OS actually assigned, so connect() dials the right one when
+        // relayPort is 0.
+        const address = this.server?.address();
+        if (address && typeof address === 'object') {
+          this.relayPort = address.port;
+        }
         log.info(`Relay server running on localhost:${this.relayPort}`);
         resolve();
       });
@@ -549,14 +556,10 @@ export async function createUsbmux(opts: Partial<SocketOptions> = {}): Promise<U
  * Connects to a device and sets up a relay service in one operation
  * @param deviceID - The device ID to connect to
  * @param port - The port on the device to connect to
- * @param relayPort - The local port to use for the relay server
+ * @param relayPort - The local port to use for the relay server. Defaults to 0 (ephemeral).
  * @returns Promise that resolves with a connected socket
  */
-export async function connectAndRelay(
-  deviceID: string | number,
-  port: number,
-  relayPort: number = 2222,
-): Promise<Socket> {
+export async function connectAndRelay(deviceID: string | number, port: number, relayPort: number = 0): Promise<Socket> {
   // Create and start the relay service
   const relay = new RelayService(deviceID, port, relayPort);
   let socket: Socket | undefined;


mikebruin:
Two follow-ups after running this change on the affected 14-device farm.

## The fix holds up in production

The patched build has been running since the farm's tunnel daemon was restarted with it:

| | before | after |
| --- | --- | --- |
| `EADDRINUSE` occurrences | thousands (single devices reached 7051 and 4029 reconnect attempts) | **0** |
| daemon log growth | ~220 lines / 10s | **0 lines / 10s** |
| tunnels established | 12 of 14, with 2 devices permanently stuck on the port | **14 of 14, 0 failed** |

The two devices that had been unrecoverable for days — both healthy and USB-connected the whole
time, both rejected by the XCUITest driver as `Unknown device or simulator UDID` — came up
immediately once the port contention was gone.

## A third issue worth its own ticket

Related to the two already listed in the description, and worth mentioning because it is what made
the recovery a two-step process rather than one.

In `scripts/tunnel-creation.mjs`, the startup loop only attaches the lifecycle watch when initial
creation succeeds:

```js
for (const device of devicesToProcess) {
  const result = await createTunnelForDevice(device);
  results.push(result);

  if (result.success) {
    attachTunnelRegistryLifecycleWatch(registryServer.getRegistry(), [result], {
      onTunnelDead: async ({udid}) => {
        await reconnectTunnelByUdid(udid);
      },
    });
    // ...
  }
}
```

A device whose tunnel is created successfully and later drops gets `onTunnelDead` →
`reconnectTunnelByUdid`. A device that **fails during initial creation** gets no watch at all, so it
is never retried — there is no path back short of a full daemon restart.

Observed directly. On the first restart with the patch applied, 2 of the 14 devices failed initial
creation because they happened to be mid-test at that moment:

```
ERR! TunnelCreation ❌ Failed to create tunnel for device <udid>: Error: Failed to receive any data
within the timeout: 5000ms - The device might be busy or unresponsive
```

Both then sat with **zero** reconnect attempts logged, absent from the tunnel registry, and rejected
by the driver — despite being healthy and responsive to lockdown within seconds. A second daemon
restart, run while the phones were idle, brought both up and produced the 14/14 above.

This compounds the no-rescan issue in the description: between them, the only remedy for a device in
either state is restarting the daemon, which tears down every working tunnel on the host. For a
device farm that is a meaningful availability cost. Attaching the lifecycle watch (or a bounded
retry) to initial failures as well as to later drops would make both states self-healing.

Happy to open this as a separate issue, or fold a fix into this PR if you would prefer to keep it
together.

src/lib/lockdown/index.ts
```diff
@@ -18,7 +18,8 @@ const deviceManagerLog = getLogger('DeviceManager');
 const LABEL = 'appium-internal';
 const DEFAULT_TIMEOUT = 5000;
 const DEFAULT_LOCKDOWN_PORT = 62078;
-const DEFAULT_RELAY_PORT = 2222;
+/** 0 lets the OS assign a free ephemeral port, so concurrent relays cannot collide. */
+const DEFAULT_RELAY_PORT = 0;
```
mykola-mokhnach, line 22: this would be a breaking change. I would rather leave the default value and document a new env variable, like APPIUM_XPC_TUNNEL_RELAY_PORT. If this variable is not set then 2222 is still used, otherwise the value of this var is used, including zero, which means a random port

mykola-mokhnach:
Also, is there any reason to run multiple servers on a single host? The idea is that there is always only one server running on the machine managing all active tunnels

mykola-mokhnach:
Based on the issue described above, wouldn't it be more logical to implement proper sync for device reconnections as well as add throttling strategy(-ies) to the reconnect logic?

mikebruin:
Both fair questions — answering together since they're related.

**On "only one server managing tunnels":** That invariant holds for `tunnel-creation.mjs` itself, but it doesn't cover this bug, because the collision isn't (only) between two instances of the daemon. Two separate things bind the relay port, and only one of them is "the tunnel server":

1. Inside a single daemon instance, `reconnectingByUdid` is keyed per-UDID, not global — so N dropped devices retry concurrently, each binding the same hardcoded port. One server, still N concurrent binders.
2. `appium-xcuitest-driver`'s `withUsbMuxLockdown` calls `createLockdownServiceByUDID` directly from the *session-creation* process during a live test — a completely separate process from the daemon, with no awareness of it. That call path was never part of "the one server managing tunnels"; it's ad hoc lockdown access that happens to share the same relay code.

So even with exactly one `tunnel-creation.mjs` running, you still get the daemon's own reconnects colliding with each other, and either of those colliding with any Appium session doing a lockdown read at the same moment.

**On sync + throttling instead:** I think that's a good improvement, but it's a different fix for a different layer, and it wouldn't close the gap above. Adding sync/throttling to `tunnel-creation.mjs`'s reconnect loop can only coordinate code the daemon owns — it has no reach into the driver's session-creation process, which is the second collision source. You'd still need *something* to stop that cross-process contention, and once that something exists, the daemon-internal collision is fixed for free too.

The reason I went with the ephemeral port rather than coordination is that nothing outside the process ever reads this port back — `start()` binds it and `connect()` immediately dials the same loopback address in the same call. It's a private rendezvous value, not a shared resource, so it doesn't actually need a fixed value or a lock; removing the constant removes the whole class of contention instead of managing it.

That said, I still think sync/throttling on the reconnect loop is worth doing — it's issue (2) in my "possibly worth separate issues" list above (unbounded 1s retry with no backoff, which is what produced the 7051/4029 attempt counts and the 60 MB log). Happy to open that as a follow-up, or take a pass at it here if you'd rather keep it in this PR.

**On the default-value change being breaking:** Understood on wanting an escape hatch, and I'm happy to add `APPIUM_XPC_TUNNEL_RELAY_PORT` regardless of how the default question resolves — that part's not in dispute.

Where I'd push back a little: if the default stays `2222` and ephemeral is opt-in via the env var, this fix only helps people who already know they have the problem — and the whole reason it took real digging to diagnose is that the *symptom* isn't `EADDRINUSE`. That error is buried in a reconnect loop or a background daemon's logs. What a user actually sees is the driver rejecting a healthy, connected device with `Unknown device or simulator UDID`, which gives no hint a port conflict is the cause. Nobody is going to set an undocumented-until-now env var to fix an error they can't see.

Since nothing outside the process ever reads or depends on the port value programmatically (it's bound and immediately dialed in the same call), I'm not sure what the breaking surface is beyond "something external assumes 2222 is open/used" — firewall rules, monitoring, manual `lsof`/`netstat` checks against that port, that kind of thing. If that's the concern, could we default to ephemeral and let `APPIUM_XPC_TUNNEL_RELAY_PORT` opt back *into* a fixed port instead of the other way around? That way the fix is protective by default, and anyone with an actual external dependency on `2222` gets an explicit, documented way to pin it.

If there's a concrete downstream case I'm not accounting for, happy to go with your original ordering instead — just want to make sure we're not shipping a fix that only works for people who've already hit and diagnosed the bug.

---
_Analysis co-authored with Claude (Anthropic)._

mykola-mokhnach:
Thanks for the detailed reply, even though I'd prefer to have it more concise. Claude tends to spit too much of nitty-gritty details making it hard to parse. I think it does make sense to randomize the port and apply this PR, although it would still only hide the original problem, which is the fact that we don't properly close the relay server on failure. https://github.com/appium/appium-ios-remotexpc/pull/287 should address that.

@harsha509 @navin772 WDYT?

mykola-mokhnach (approved):

harsha509 (approved):
HI @mykola-mokhnach ,

Agreed on #287, that is a real leak. service is still undefined when connect() is what failed, so the relay server was never stopped and held 2222 for the rest of the process. That is what turned one collision into the permanent 7000 retry state.

I would not say this PR only hides it, though, each relay holds the port for the whole lifetime of its LockdownService, not just for a moment. So two overlapping calls collide even when nothing leaks, which is what Mike's two concurrent call repro shows. And `withUsbMuxLockdown` in the `xcuitest driver` opens a relay from a separate process, so #287 cannot reach that one at all.

#287 fixes what makes it permanent. This one fixes what makes it happen.

LGTM!

harsha509:
Thank you for the fix @mikebruin !

PR #287, opened 2026-08-07
fix: stop leaked relay server when LockdownServiceFactory.createByUDID fails

mykola-mokhnach:
If relay.connect() failed after relay.start() already succeeded, cleanup routed through service?.stopRelayService(...), which is a no-op since service is only assigned after a successful connect(). The already- listening relay server was never stopped, leaking its bound port for the rest of the process's life and breaking subsequent createLockdownServiceByUDID calls with EADDRINUSE.

See https://github.com/appium/appium-ios-remotexpc/pull/286 for more details


harsha509 (approved):
Thank you @mykola-mokhnach !

navin772 (approved):

PR #288, opened 2026-08-07
chore: replace npm-run-all2 with a plain npm script chain

mykola-mokhnach:
The build script only ever ran two steps in sequence, so run-s added an extra dependency for no real benefit.

Also bumped node-devicectl after ESM update


harsha509 (approved):
LGTM!

navin772 (approved):

PR #289, opened 2026-08-11
fix(display): surface screen-recording write errors instead of crashing

harsha509:
## What this resolves

`recordScreenToFile` and `recordScreenAndAudioToFiles` create their Annex-B output with `createWriteStream` and never attach an `'error'` listener. A stream error therefore reaches the process as an **uncaught exception** rather than a rejected promise.

The reason this matters is not tidiness — it is that **the caller cannot defend against it**. The event bypasses the promise chain entirely, so a consumer doing everything right is still killed:

```js
try {
  await recordScreenToFile(service, outputPath);
} catch (e) {
  // never runs — the process is already gone
}
```

Verified on Node 24:

```
caller wraps the call in try/catch and awaits it...
>>> UNCAUGHT EXCEPTION escaped the caller try/catch: ENOENT
```

The realistic trigger is ordinary, not exotic: an `outputPath` whose parent directory does not exist. A full disk or an unplugged volume mid-recording does the same thing. `end(callback)` is not a substitute — the callback *does* receive the error, but `'error'` is still emitted separately and still terminates the process.

The audio half of this same feature already guards against exactly this in `M4aFileWriter` (`m4a-writer.ts:336-342`, which spells out why it matters for unattended recordings). The video half did not.

## The change

`AnnexBFileWriter` gives the video track the same guarantee: it holds the first stream error and re-throws it from the next `write()` or `close()`, so a failed recording surfaces as a rejection the caller can handle. It replaces the inline `createWriteStream` + `write`/`once('drain')` + `end()` blocks at both call sites, so the net new code is about +9 lines.

Two details differ from `M4aFileWriter` deliberately:

- **It does not await `'open'`.** Doing so would reject while the capture is already streaming, stranding it with no handle to stop it. An open failure surfaces from the first `write()` instead.
- **`close()` reports the held error in preference to ending an already-failed stream**, which answers with a generic `ERR_STREAM_DESTROYED` and buries the cause. That is the common path, since the recorders write nothing until the first keyframe arrives.

Backpressure, output bytes and public API are unchanged — `AnnexBFileWriter` is module-internal and is not exported from `src/index.ts`.

## Verification

Unit tests added to `test/unit/display/video/screen-stream-capture.spec.ts` covering both crash paths plus the write and backpressure behaviour. They were mutation-tested: stripping the `'error'` listener from the compiled output fails **exactly** the two crash-guard tests and passes the other three, so they lock in this change and nothing more.

```
✔ writes every chunk through to the file
✔ writes a file larger than the stream buffer, so backpressure is exercised
✔ reports a bad output path as a rejection rather than an uncaught error
✔ reports a bad output path from close() when no write ever ran
✔ surfaces a write failure through the next call
```

| Check | Result |
| --- | --- |
| `tsc --noEmit` | clean |
| `oxlint` | clean |
| `oxfmt --check` | clean |
| Display unit tests | 164/164 pass |

Not verified end-to-end against hardware: `recordScreenToFile` / `recordScreenAndAudioToFiles` need an iOS 27 device (`npm run test:display`). What is verified is the writer that was the defect.

## Scope

Deliberately limited to this one issue. Three related findings turned up while reading the surrounding code and are **not** addressed here, to keep the change reviewable:

1. `M4aFileWriter.create` can reject after the captures are already running in `recordScreenAndAudioToFiles` and `recordAudioToFile`, stranding a live device stream with an unbounded RTP queue.
2. In `recordScreenAndAudioToFiles`, `audioOut.close()` precedes `videoOut.close()`, so a throwing audio close skips the video close.
3. `parseRtpPacket` ignores the RTP padding bit (spec completeness; no evidence the device sets it).

Happy to open a follow-up for 1 and 2 if reviewers would rather see them fixed together.

### DIFF
--- a/src/services/ios/display/video/screen-stream-capture.ts
+++ b/src/services/ios/display/video/screen-stream-capture.ts
@@ -191,6 +191,69 @@ export class ScreenStreamCapture {
   }
 }
 
+/**
+ * The Annex-B output file, with the write stream's `'error'` event held instead
+ * of left to reach the process.
+ *
+ * A recording runs unattended for minutes and the stream never escapes the
+ * function that owns it, so a caller cannot attach a listener of its own.
+ * Without one here, a mistyped path or a write that fails after the fact (a full
+ * disk, an unplugged volume) arrives as an uncaught `'error'` event and takes
+ * the process down — `end(callback)` does not help, since the callback receives
+ * the error *and* the event is still emitted. Holding the first error and
+ * re-throwing it from the next `write` or `close` turns that into a rejection
+ * the caller can see, the guarantee {@link M4aFileWriter} already gives the
+ * audio track.
+ *
+ * Opening is deliberately not awaited, unlike `M4aFileWriter.create`: the
+ * capture is already streaming by the time the file is created, so a rejection
+ * before the caller holds the writer would strand it. An open failure surfaces
+ * from the first {@link write} instead.
+ */
+export class AnnexBFileWriter {
+  private readonly stream: ReturnType<typeof createWriteStream>;
+  /** First stream error seen, re-thrown from the next `write` or `close`. */
+  private streamError: Error | undefined;
+
+  /** @param path Destination file path; truncated if it exists. */
+  constructor(path: string) {
+    this.stream = createWriteStream(path);
+    this.stream.on('error', (error: Error) => {
+      this.streamError ??= error;
+    });
+  }
+
+  /**
+   * Appends one chunk, resolving once the stream has room for more so an
+   * `await` per chunk applies backpressure.
+   */
+  async write(chunk: Buffer): Promise<void> {
+    if (this.streamError) {
+      throw this.streamError;
+    }
+    if (!this.stream.write(chunk)) {
+      await once(this.stream, 'drain');
+    }
+    if (this.streamError) {
+      throw this.streamError;
+    }
+  }
+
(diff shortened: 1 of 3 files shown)


src/services/ios/display/video/screen-stream-capture.ts
```diff
@@ -191,6 +191,69 @@ export class ScreenStreamCapture {
...
+ */
+export class AnnexBFileWriter {
+  private readonly stream: ReturnType<typeof createWriteStream>;
+  /** First stream error seen, re-thrown from the next `write` or `close`. */
+  private streamError: Error | undefined;
+
+  /** @param path Destination file path; truncated if it exists. */
+  constructor(path: string) {
+    this.stream = createWriteStream(path);
+    this.stream.on('error', (error: Error) => {
```
mykola-mokhnach, line 221: once?

mykola-mokhnach (approved):

src/services/ios/display/video/screen-stream-capture.ts
harsha509: I kept `.on` because that suppression is stream behaviour, not `EventEmitter`
behaviour. If this ever wrapped a different Writable, a second `'error'` would
go uncaught, which is the exact thing this class prevents. It also matches
`M4aFileWriter`.

Happy to switch to `.once` if you prefer, with `??=` becoming `=`.

PR #293, opened 2026-08-13
feat(accessibility): add AccessibilityAuditService for on-device audits and element inspection

navin772:
Adds `com.apple.accessibility.axAuditDaemon.remoteserver` - the DTX backend behind Xcode's Accessibility Inspector.

### Public API

`Services.startAccessibilityAuditService(udid)` → `AccessibilityAuditService`

| Method | Returns |
| --- | --- |
| `getApiVersion(options?)` | `number` |
| `getCapabilities(options?)` | `string[]` — the 45 selectors the daemon implements |
| `getSupportedAuditTypes(options?)` | `string[]` — the `testType*` catalogue |
| `getAccessibilitySettings(options?)` | `AxDeviceSetting[]` — settings with current values |
| `runAudit(auditTypes, options?)` | `AxAuditIssue[]` — issues on the current screen |
| `getFocusedElement(options?)` | `AxInspectedElement` — the full inspector panel |
| `observeFocusedElement(listener, options?)` | `() => void` — focus-change stream |
| `getElementAttributeValue(element, attribute, options?)` | `unknown` |
| `getSpecialElement(index, options?)` | `AxElement \| undefined` |
| `close()` | — |

`RunAuditOptions`: `targetPid`, `timeoutMs`, `onLog`
`InspectOptions`: `timeoutMs`, `showVisuals`

Note: there are some more methods which would be added in a separate PR.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -59,6 +59,11 @@ export type {
   ListAppsOptions,
 } from './services/ios/app-service/index.js';
 export {PasteboardService} from './services/ios/pasteboard/index.js';
+export {AccessibilityAuditService} from './services/ios/accessibility-audit/index.js';
+export type {AxDeviceSetting} from './services/ios/accessibility-audit/index.js';
+export {AxAuditDtxTransport} from './services/ios/accessibility-audit/dtx-transport.js';
+export type {InvokeOptions as AxInvokeOptions} from './services/ios/accessibility-audit/dtx-transport.js';
+export {AX_OBJECT_TYPE, deserializeAxObject} from './services/ios/accessibility-audit/ax-deserialize.js';
 export {CoreDeviceInfoService} from './services/ios/device-info/index.js';
 export type {
   CoreDeviceAttributes,
--- a/src/index.ts
+++ b/src/index.ts
@@ -65,7 +65,12 @@ export {AxAuditDtxTransport} from './services/ios/accessibility-audit/dtx-transp
 export type {InvokeOptions as AxInvokeOptions} from './services/ios/accessibility-audit/dtx-transport.js';
 export {AX_OBJECT_TYPE, deserializeAxObject} from './services/ios/accessibility-audit/ax-deserialize.js';
 export {AxPoint} from './services/ios/accessibility-audit/ax-values.js';
-export {serializeAxElement, toAxElement, toInspectedElement} from './services/ios/accessibility-audit/ax-element.js';
+export {
+  serializeAxAttribute,
+  serializeAxElement,
+  toAxElement,
+  toInspectedElement,
+} from './services/ios/accessibility-audit/ax-element.js';
 export type {
   AxElement,
   AxElementAttribute,
--- a/src/index.ts
+++ b/src/index.ts
@@ -64,6 +64,15 @@ export type {AxDeviceSetting} from './services/ios/accessibility-audit/index.js'
 export {AxAuditDtxTransport} from './services/ios/accessibility-audit/dtx-transport.js';
 export type {InvokeOptions as AxInvokeOptions} from './services/ios/accessibility-audit/dtx-transport.js';
 export {AX_OBJECT_TYPE, deserializeAxObject} from './services/ios/accessibility-audit/ax-deserialize.js';
+export {AxPoint} from './services/ios/accessibility-audit/ax-values.js';
+export {serializeAxElement, toAxElement, toInspectedElement} from './services/ios/accessibility-audit/ax-element.js';
+export type {
+  AxElement,
+  AxElementAttribute,
+  AxInspectedElement,
+  AxInspectorSection,
+} from './services/ios/accessibility-audit/ax-element.js';
+export type {InspectOptions, RunAuditOptions, AxAuditIssue} from './services/ios/accessibility-audit/index.js';
 export {CoreDeviceInfoService} from './services/ios/device-info/index.js';
 export type {
   CoreDeviceAttributes,
--- a/src/services/ios/accessibility-audit/ax-deserialize.ts
+++ b/src/services/ios/accessibility-audit/ax-deserialize.ts
@@ -1,0 +1,54 @@
+/**
+ * The accessibility audit daemon wraps every value it returns in a recursive
+ * `{Value, ObjectType}` envelope. `ObjectType: "passthrough"` is a plain boxed
+ * value; any other `ObjectType` names a typed object (e.g.
+ * `AXAuditDeviceSetting_v1`) whose `Value` is a dictionary of fields.
+ */
+
+/** Key under which a typed (non-passthrough) object records its `ObjectType`. */
+export const AX_OBJECT_TYPE = '__axObjectType';
+
+/** A decoded typed object: its fields plus the {@link AX_OBJECT_TYPE} tag. */
+export type AxTypedObject = Record<string, unknown> & {[AX_OBJECT_TYPE]: string};
+
+function isEnvelope(value: unknown): value is {Value: unknown; ObjectType: string} {
+  return (
+    typeof value === 'object' &&
+    value !== null &&
+    !Array.isArray(value) &&
+    'ObjectType' in value &&
+    typeof (value as {ObjectType: unknown}).ObjectType === 'string'
+  );
+}
+
+/**
+ * Recursively unwraps the daemon's serialized-object envelopes.
+ *
+ * @param value A value decoded from an NSKeyedArchiver reply.
+ */
+export function deserializeAxObject(value: unknown): unknown {
+  if (Array.isArray(value)) {
+    return value.map(deserializeAxObject);
+  }
+  if (!isEnvelope(value)) {
+    if (typeof value === 'object' && value !== null) {
+      // A plain dictionary with no ObjectType: deserialize each field.
+      const out: Record<string, unknown> = {};
+      for (const [key, inner] of Object.entries(value)) {
+        out[key] = deserializeAxObject(inner);
+      }
+      return out;
+    }
+    return value;
+  }
+
+  const inner = deserializeAxObject(value.Value);
+  if (value.ObjectType === 'passthrough') {
+    return inner;
+  }
+  // A typed object. Spread its fields (when it has them) and tag the type.
+  if (typeof inner === 'object' && inner !== null && !Array.isArray(inner)) {
+    return {...(inner as Record<string, unknown>), [AX_OBJECT_TYPE]: value.ObjectType};
+  }
+  return {value: inner, [AX_OBJECT_TYPE]: value.ObjectType};
+}
--- a/src/services/ios/accessibility-audit/ax-deserialize.ts
+++ b/src/services/ios/accessibility-audit/ax-deserialize.ts
@@ -4,6 +4,7 @@
  * value; any other `ObjectType` names a typed object (e.g.
  * `AXAuditDeviceSetting_v1`) whose `Value` is a dictionary of fields.
  */
+import {util} from '@appium/support';
 
 /** Key under which a typed (non-passthrough) object records its `ObjectType`. */
 export const AX_OBJECT_TYPE = '__axObjectType';
@@ -13,9 +14,7 @@ export type AxTypedObject = Record<string, unknown> & {[AX_OBJECT_TYPE]: string}
 
 function isEnvelope(value: unknown): value is {Value: unknown; ObjectType: string} {
   return (
-    typeof value === 'object' &&
-    value !== null &&
-    !Array.isArray(value) &&
+    util.isPlainObject(value) &&
     'ObjectType' in value &&
     typeof (value as {ObjectType: unknown}).ObjectType === 'string'
   );
@@ -31,7 +30,7 @@ export function deserializeAxObject(value: unknown): unknown {
     return value.map(deserializeAxObject);
   }
   if (!isEnvelope(value)) {
-    if (typeof value === 'object' && value !== null) {
+    if (util.isPlainObject(value)) {
       // A plain dictionary with no ObjectType: deserialize each field.
       const out: Record<string, unknown> = {};
       for (const [key, inner] of Object.entries(value)) {
@@ -47,7 +46,7 @@ export function deserializeAxObject(value: unknown): unknown {
     return inner;
   }
   // A typed object. Spread its fields (when it has them) and tag the type.
-  if (typeof inner === 'object' && inner !== null && !Array.isArray(inner)) {
+  if (util.isPlainObject(inner)) {
     return {...(inner as Record<string, unknown>), [AX_OBJECT_TYPE]: value.ObjectType};
   }
   return {value: inner, [AX_OBJECT_TYPE]: value.ObjectType};
--- a/src/services/ios/accessibility-audit/ax-element.ts
+++ b/src/services/ios/accessibility-audit/ax-element.ts
@@ -1,3 +1,5 @@
+import {util} from '@appium/support';
+
 import {AX_OBJECT_TYPE} from './ax-deserialize.js';
 
 /**
@@ -73,17 +75,22 @@ function toBuffer(value: unknown): Buffer | undefined {
   return undefined;
 }
 
-/** Parses a deserialized `AXAuditElement_v1`. */
+/**
+ * Parses a deserialized `AXAuditElement_v1`.
+ *
+ * The `_v1` suffixes are the daemon's own wire keys, not our assumption. A
+ * future shape would carry different keys, so this returns `undefined` rather
+ * than misreading one.
+ */
 export function toAxElement(value: unknown): AxElement | undefined {
-  if (typeof value !== 'object' || value === null) {
+  if (!util.isPlainObject(value)) {
     return undefined;
   }
   const fields = value as Record<string, unknown>;
   const platformValue = fields.PlatformElementValue_v1;
-  const container =
-    typeof platformValue === 'object' && platformValue !== null
-      ? ((platformValue as Record<string, unknown>)['NS.data'] ?? platformValue)
-      : undefined;
+  const container = util.isPlainObject(platformValue)
+    ? ((platformValue as Record<string, unknown>)['NS.data'] ?? platformValue)
+    : undefined;
   const platformElement = toBuffer(container);
   if (!platformElement) {
     return undefined;
@@ -113,7 +120,7 @@ export function serializeAxElement(element: AxElement): Record<string, unknown> 
 }
 
 function toAttribute(value: unknown): AxElementAttribute | undefined {
-  if (typeof value !== 'object' || value === null) {
+  if (!util.isPlainObject(value)) {
     return undefined;
   }
   const fields = value as Record<string, unknown>;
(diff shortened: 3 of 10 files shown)


src/services/ios/accessibility-audit/ax-element.ts
```diff
@@ -0,0 +1,183 @@
...
+  const container =
+    typeof platformValue === 'object' && platformValue !== null
+      ? ((platformValue as Record<string, unknown>)['NS.data'] ?? platformValue)
+      : undefined;
+  const platformElement = toBuffer(container);
+  if (!platformElement) {
+    return undefined;
+  }
+  return {
+    platformElement,
```
mykola-mokhnach, line 99: how do we know if this element shape is v1? is it going to break if Apple introduces v2?

src/services/ios/accessibility-audit/ax-element.ts
```diff
@@ -0,0 +1,183 @@
...
+    value.AccessibilityIdentifier_v1 = {ObjectType: 'passthrough', Value: element.accessibilityIdentifier};
+  }
+  return {
+    ObjectType: 'AXAuditElement_v1',
+    Value: {ObjectType: 'passthrough', Value: value},
+  };
+}
+
+function toAttribute(value: unknown): AxElementAttribute | undefined {
+  if (typeof value !== 'object' || value === null) {
```
mykola-mokhnach: consider using util.isPlainObject check from appium support lib to verify if a value is a plain object/Record

src/services/ios/accessibility-audit/ax-element.ts
```diff
@@ -0,0 +1,183 @@
...
+      out[key] = value;
+    }
+  }
+  return out;
+}
+
+/** Rebuilds an attribute descriptor for the wire. */
+export function serializeAxAttribute(attribute: AxElementAttribute): Record<string, unknown> {
+  const value: Record<string, unknown> = {};
+  for (const [key, inner] of Object.entries(attribute.raw)) {
```
mykola-mokhnach: nit: this could use reduce

src/services/ios/accessibility-audit/dtx-transport.ts
```diff
@@ -0,0 +1,529 @@
...
+
+  private async open(): Promise<void> {
+    // A raw socket on purpose: this protocol interleaves length-prefixed plists
+    // and raw DTX on one stream, so it cannot share ServiceConnection's plist
+    // pipeline (which would consume the checkin reply before the DTX reader sees
+    // it). Only the address resolution is reused.
+    const [host, port] = await this.resolveServiceAddress(AxAuditDtxTransport.RSD_SERVICE_NAME);
+    const socket = await new Promise<net.Socket>((resolve, reject) => {
+      const created = net.createConnection({host, port, family: 6}, () => {
+        created.setKeepAlive(true);
```
mykola-mokhnach, line 122: also nodelay?

src/services/ios/accessibility-audit/dtx-transport.ts
```diff
@@ -0,0 +1,529 @@
...
+    });
+    const body = Buffer.from(String(requestXml), 'utf8');
+    const framed = Buffer.alloc(4 + body.length);
+    framed.writeUInt32BE(body.length, 0);
+    body.copy(framed, 4);
+    this.write(framed);
+
+    // Plists are length-prefixed (4-byte big-endian); the DTX stream is not, so
+    // the magic marks the boundary.
+    for (;;) {
```
mykola-mokhnach, line 230: Consider making the loop limited. If something is wrong in the protocol it might loop forever

src/services/ios/accessibility-audit/dtx-transport.ts
```diff
@@ -0,0 +1,529 @@
...
+      const head = await this.peek(4);
+      if (head.equals(DTX_MAGIC)) {
+        return;
+      }
+      const length = head.readUInt32BE(0);
+      await this.readExact(4);
+      const plist = await this.readExact(length);
+      const request = plist.toString('utf8').match(/<key>Request<\/key>\s*<string>([^<]*)</);
+      log.debug(`Checkin plist consumed: ${request ? request[1] : '(unrecognized)'}`);
```
mykola-mokhnach: this might get too noisy if multiple loop iterations are executed

src/services/ios/accessibility-audit/dtx-transport.ts
```diff
@@ -0,0 +1,529 @@
...
+   * Reassembles one complete DTX message from the socket, handling multi-fragment
+   * messages the same way {@link DVTSecureSocketProxyService} does.
+   */
+  private async recvMessage(): Promise<{
+    channel: number;
+    identifier: number;
+    conversationIndex: number;
+    payload: Buffer;
+  }> {
+    for (;;) {
```
mykola-mokhnach: same as above. put some threshold to guarantee killing of the loop before it stalls

src/services/ios/accessibility-audit/dtx-transport.ts
```diff
@@ -0,0 +1,529 @@
...
+    const objectData =
+      objectSize > 0 ? body.subarray(payloadHeader.auxiliaryLength, payloadHeader.auxiliaryLength + objectSize) : null;
+    const object = objectData ? decodeNSKeyedArchiver(parseBinaryPlist(objectData)) : null;
+    return {object, args};
+  }
+
+  /** The single read loop; dispatches replies to callers and inbound calls to waiters. */
+  private async readLoop(): Promise<void> {
+    try {
+      for (;;) {
```
mykola-mokhnach, line 314: where does this loop is supposed to break?

src/services/ios/accessibility-audit/dtx-transport.ts
```diff
@@ -0,0 +1,529 @@
...
+  const values: unknown[] = [];
+  let offset = 16;
+  while (offset + 8 <= end) {
+    // Each item is prefixed by an empty-dictionary marker then a type word.
+    if (buffer.readUInt32LE(offset) === DTX_CONSTANTS.EMPTY_DICTIONARY) {
+      offset += 4;
+    }
+    const type = buffer.readUInt32LE(offset);
+    offset += 4;
+    if (type === DTX_CONSTANTS.AUX_TYPE_OBJECT) {
```
mykola-mokhnach, line 552: nit: maybe would be easier to read if used switch

src/services/ios/accessibility-audit/dtx-transport.ts
```diff
@@ -0,0 +1,529 @@
...
+  if (values.length === 0) {
+    return Buffer.alloc(0);
+  }
+  const parts: Buffer[] = [];
+  for (const {type, value} of values) {
+    const marker = Buffer.alloc(8);
+    marker.writeUInt32LE(DTX_CONSTANTS.EMPTY_DICTIONARY, 0);
+    marker.writeUInt32LE(type, 4);
+    parts.push(marker);
+    if (type === DTX_CONSTANTS.AUX_TYPE_OBJECT) {
```
mykola-mokhnach: same as above

src/services/ios/accessibility-audit/index.ts
```diff
@@ -0,0 +1,370 @@
...
+   *
+   * @param auditTypes Audit types to run, from {@link getSupportedAuditTypes}.
+   * @param options Timeout for the completion callback.
+   */
+  async runAudit(auditTypes: string[], options: RunAuditOptions = {}): Promise<AxAuditIssue[]> {
+    // Issues are streamed one per `hostFoundAuditIssue:` call and the
+    // completion callback carries no arguments. Older releases are reported to
+    // return them in the completion instead, so both are collected and the
+    // streamed set wins when present.
+    const issues: AxAuditIssue[] = [];
```
mykola-mokhnach, line 144: should only one instance of this method be executed at once or many are also fine?

src/services/ios/accessibility-audit/index.ts
```diff
@@ -0,0 +1,370 @@
...
+   * Subscribes to focus changes, delivering an inspector panel each time the
+   * device's accessibility focus moves.
+   *
+   * Monitoring stays armed until the returned function is called.
+   *
+   * @param listener Receives each pushed element.
+   * @param options Whether to draw the on-device highlight.
+   */
+  observeFocusedElement(
+    listener: (element: AxInspectedElement) => void,
```
mykola-mokhnach, line 235: should listener be always a synchronous callback?

src/services/ios/accessibility-audit/index.ts
```diff
@@ -0,0 +1,370 @@
...
+    options: {showVisuals?: boolean} = {},
+  ): () => void {
+    const stop = this.transport.onInbound('hostInspectorCurrentElementChanged:', (args) => {
+      listener(toInspectedElement(deserializeAxObject(args[0])));
```
mykola-mokhnach, line 239: what if it throws an error?

src/services/ios/accessibility-audit/index.ts
```diff
@@ -0,0 +1,370 @@
...
+    });
+    this.setMonitoredEventType(MONITORED_EVENT_FOCUS);
+    if (options.showVisuals) {
+      this.setShowVisuals(true);
+    }
+    return () => {
+      stop();
```
mykola-mokhnach, line 252: do we cleanup all pending listeners when the service is closed?

src/services/ios/accessibility-audit/dtx-transport.ts
```diff
@@ -0,0 +1,529 @@
...
+  private onData(chunk: Buffer): void {
+    this.readBuffer = Buffer.concat([this.readBuffer, chunk]);
+    const waiter = this.readWaiter;
+    if (waiter) {
+      this.readWaiter = null;
+      waiter();
+    }
+  }
+
+  private onClose(error: Error): void {
```
harsha509, line 157: When the connection closes, `pendingReplies` are rejected but `inboundWaiters` are not. Any pending `waitForInbound` promise (used by runAudit and getFocusedElement) will hang until its own timeout fires instead of rejecting immediately. 
we ned to `reject/clear` inboundWaiters here

src/services/ios/accessibility-audit/dtx-transport.ts
```diff
@@ -0,0 +1,529 @@
...
+        if (waiters && waiters.length > 0) {
+          for (const waiter of waiters.splice(0)) {
+            waiter(args);
+          }
+        } else if (!listeners || listeners.length === 0) {
+          log.debug(`Inbound ${selector || 'message'} dropped (${args.length} arg(s))`);
+        }
+      }
+    } catch (error) {
+      this.onClose(error instanceof Error ? error : new Error(String(error)));
```
harsha509, line 325: The `try/catch` wraps the entire read loop, so if a single inbound frame fails to decode in `decodePayload` the loop exits and the whole connection is torn down. 

Consider catching decode errors per message (log + skip) so one malformed frame doesnt kill the session.

src/services/ios/accessibility-audit/index.ts
```diff
@@ -0,0 +1,370 @@
...
+        'hostDeviceDidCompleteAuditCategoriesWithAuditIssues:',
+        options.timeoutMs,
+      );
+      const aux = new MessageAux();
+      aux.appendObj(auditTypes);
+      this.transport.invokeOneway('deviceBeginAuditTypes:', aux);
+      const completionArgs = await completion;
+      // The completion call can land marginally before the last issue is read
+      // off the socket, so let the queue drain.
+      await new Promise((resolve) => setTimeout(resolve, 250));
```
harsha509: This fixed `250ms` sleep means any `hostFoundAuditIssue:` arriving later is silently dropped from the results. 

Can we replace this with a deterministic wait ?

src/services/ios/accessibility-audit/dtx-transport.ts
```diff
@@ -0,0 +1,529 @@
...
+  }
+
+  private async open(): Promise<void> {
+    // A raw socket on purpose: this protocol interleaves length-prefixed plists
+    // and raw DTX on one stream, so it cannot share ServiceConnection's plist
+    // pipeline (which would consume the checkin reply before the DTX reader sees
+    // it). Only the address resolution is reused.
+    const [host, port] = await this.resolveServiceAddress(AxAuditDtxTransport.RSD_SERVICE_NAME);
+    const socket = await new Promise<net.Socket>((resolve, reject) => {
+      const created = net.createConnection({host, port, family: 6}, () => {
```
harsha509: Neither the socket connect nor the checkin loop below has a timeout.

harsha509: only ipv6 ?

src/services/ios/accessibility-audit/index.ts
```diff
@@ -0,0 +1,370 @@
...
+    this.setMonitoredEventType(MONITORED_EVENT_FOCUS);
+    if (options.showVisuals) {
+      this.setShowVisuals(true);
+    }
+    return () => {
+      stop();
+      if (options.showVisuals) {
+        this.setShowVisuals(false);
+      }
+      this.setMonitoredEventType(MONITORED_EVENT_OFF);
```
harsha509: Unsubscribing one observer disarms monitoring unconditionally....!

src/index.ts
```diff
@@ -59,6 +59,20 @@ export type {
   ListAppsOptions,
 } from './services/ios/app-service/index.js';
 export {PasteboardService} from './services/ios/pasteboard/index.js';
+export {AccessibilityAuditService} from './services/ios/accessibility-audit/index.js';
+export type {AxDeviceSetting} from './services/ios/accessibility-audit/index.js';
+export {AxAuditDtxTransport} from './services/ios/accessibility-audit/dtx-transport.js';
+export type {InvokeOptions as AxInvokeOptions} from './services/ios/accessibility-audit/dtx-transport.js';
+export {AX_OBJECT_TYPE, deserializeAxObject} from './services/ios/accessibility-audit/ax-deserialize.js';
+export {AxPoint} from './services/ios/accessibility-audit/ax-values.js';
+export {serializeAxElement, toAxElement, toInspectedElement} from './services/ios/accessibility-audit/ax-element.js';
```
harsha509: `export serializeAxAttribute` ?

harsha509 (changes requested):

src/services/ios/accessibility-audit/ax-element.ts
navin772: They are tagged as v1 - `__axObjectType: "AXAuditElement_v1"`.
A v2 shape would carry different keys, so `toAxElement` returns undefined.

src/services/ios/accessibility-audit/dtx-transport.ts
navin772: Added `setNoDelay(true)`

src/services/ios/accessibility-audit/dtx-transport.ts
navin772: capped at 8 plists

src/services/ios/accessibility-audit/dtx-transport.ts
navin772: Fixed, now it collects and logs only once

src/services/ios/accessibility-audit/dtx-transport.ts
navin772: capped at 65536 fragments

src/services/ios/accessibility-audit/dtx-transport.ts
navin772: Only when `recvMessage` throws, which happens once the socket closes and peek rejects with the close error.

src/services/ios/accessibility-audit/index.ts
navin772: One per service instance. For concurrent calls, it shares only 1 inbound stream so it gets mixed

src/services/ios/accessibility-audit/index.ts
navin772: Fixed, listener calls are now individually guarded

src/services/ios/accessibility-audit/index.ts
navin772: yes

src/services/ios/accessibility-audit/index.ts
navin772: fixed, now it clears both waiter and listener maps

src/services/ios/accessibility-audit/dtx-transport.ts
navin772: fixed

src/services/ios/accessibility-audit/index.ts
navin772: removed the 250ms wait, we don't need it, tested and it works fine

src/services/ios/accessibility-audit/dtx-transport.ts
navin772: Added a 30s connect timeout and handshake deadline

navin772: Dropped `family: 6`

src/index.ts
navin772: added

src/services/ios/accessibility-audit/index.ts
navin772: Fixed, also updated `getFocusedElement`

harsha509 (approved):
LGTM!

src/services/ios/accessibility-audit/ax-deserialize.ts
```diff
@@ -0,0 +1,54 @@
...
+
+/** Key under which a typed (non-passthrough) object records its `ObjectType`. */
+export const AX_OBJECT_TYPE = '__axObjectType';
+
+/** A decoded typed object: its fields plus the {@link AX_OBJECT_TYPE} tag. */
+export type AxTypedObject = Record<string, unknown> & {[AX_OBJECT_TYPE]: string};
+
+function isEnvelope(value: unknown): value is {Value: unknown; ObjectType: string} {
+  return (
+    typeof value === 'object' &&
```
mykola-mokhnach: use isPlainObject helper from support util

src/services/ios/accessibility-audit/ax-deserialize.ts
```diff
@@ -0,0 +1,54 @@
...
+    }
+    return value;
+  }
+
+  const inner = deserializeAxObject(value.Value);
+  if (value.ObjectType === 'passthrough') {
+    return inner;
+  }
+  // A typed object. Spread its fields (when it has them) and tag the type.
+  if (typeof inner === 'object' && inner !== null && !Array.isArray(inner)) {
```
mykola-mokhnach: same as above

src/services/ios/accessibility-audit/dtx-transport.ts
```diff
@@ -0,0 +1,628 @@
...
+  // capacity rather than identifying the format. Requiring an exact match
+  // silently discarded every audit issue. The declared item length is validated
+  // instead, and the item loop bails on anything it does not recognise.
+  const itemsLength = Number(buffer.readBigUInt64LE(8));
+  if (itemsLength <= 0 || itemsLength > buffer.length) {
+    return [];
+  }
+  const end = Math.min(16 + itemsLength, buffer.length);
+  const values: unknown[] = [];
+  let offset = 16;
```
mykola-mokhnach: it would be nice to move magic numbers into constants

src/services/ios/accessibility-audit/dtx-transport.ts
```diff
@@ -0,0 +1,628 @@
...
+  }
+  return values;
+}
+
+/** Rejects if `promise` has not settled within `timeoutMs`. */
+async function withDeadline<T>(promise: Promise<T>, timeoutMs: number, what: string): Promise<T> {
+  let timer: NodeJS.Timeout | undefined;
+  try {
+    return await Promise.race([
+      promise,
```
mykola-mokhnach, line 589: is it ok if the promise keeps running after the race? what if it throws an exception later?

src/services/ios/accessibility-audit/index.ts
mykola-mokhnach: shall we then add some guards to avoid this mixing?

mykola-mokhnach (approved):

src/services/ios/accessibility-audit/index.ts
navin772: Yes, we could, will update the PR

src/services/ios/accessibility-audit/dtx-transport.ts
navin772: Yes, its bounded and now been logged

PR #299, opened 2026-08-21
fix: surface webinspector receive errors to the message iterator

harsha509:
## Summary

Receive failures in the WebInspector listen loop were emitted on the unlistened `error` event, which crashed the process. They are now emitted as `errorMessage`, the iterator stops, and the error is rethrown to the `listenMessage` consumer.

## Changes

- Emit `errorMessage` instead of `error` on receive failure
- `listenMessage` stops on `errorMessage` and rethrows the error to the caller
- Clean up the `errorMessage` listener on iterator exit and in `stopListening`
- Add a unit test covering the error propagation

### DIFF
--- a/src/services/ios/webinspector/index.ts
+++ b/src/services/ios/webinspector/index.ts
@@ -105,8 +105,15 @@ export class WebInspectorService extends BaseService {
       }
     };
 
+    let receiveError: unknown = null;
+    const errorHandler = (error: unknown) => {
+      receiveError = error;
+      stopHandler();
+    };
+
     this.messageEmitter.on('message', messageHandler);
     this.messageEmitter.once('stop', stopHandler);
+    this.messageEmitter.once('errorMessage', errorHandler);
 
     try {
       while (!stopped) {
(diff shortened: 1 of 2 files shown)


mykola-mokhnach (approved):

src/services/ios/webinspector/index.ts
```diff
@@ -382,9 +395,9 @@ export class WebInspectorService extends BaseService {
               continue;
             }
 
-            // For other errors, log and exit
+            // For other errors, log and end the iterator
             log.error('Error receiving message:', error);
-            this.messageEmitter.emit('error', error);
```
mykola-mokhnach, line 387: is it a breaking change?

harsha509: no Mykola, the emitter is private here..

kbsoso9999-maker: d5df32b489675031f446bbd2123dddbec9b21a8d

kbsoso9999-maker:
- [ ]

PR #300, opened 2026-08-23
fix: serialize DVT socket reads through a single receive pump

harsha509:
## Problem

DVTSecureSocketProxyService read the socket by attaching a once('data') listener inside readExact for every call. Two concurrent receivers both consumed the same chunk, which duplicated bytes into the shared read buffer and desynchronized the DTX stream. Chunks that arrived while no read was pending were dropped, and a receiver blocked on a dead socket could hang until an unrelated socket event.

## Fix

- Attach data, error and close listeners once per connection and route every read through a single pump, so only one consumer touches the byte stream
- Parse frames sequentially and resolve waiters per channel, which lets concurrent receivers on different channels proceed without corrupting framing
- Validate the DTX header magic and fail pending receivers on a desynchronized stream instead of misparsing it
- Deliver replies that were fully received before the socket closed instead of rejecting them
- Queue incoming chunks and merge them lazily, and pause the socket while nothing is reading so TCP backpressure throttles the device
- Reject all pending receivers when the pump fails so no receiver is left hanging
- Add unit coverage for the receive path

### DIFF
--- a/src/services/ios/dvt/index.ts
+++ b/src/services/ios/dvt/index.ts
@@ -22,6 +22,13 @@ const log = getLogger('DVTSecureSocketProxyService');
 
 const MIN_ERROR_DESCRIPTION_LENGTH = 20;
 
+interface ChannelWaiter {
+  resolve: (message: Buffer) => void;
+  reject: (err: Error) => void;
+  signal?: AbortSignal;
+  onAbort?: () => void;
+}
+
 /**
  * DVTSecureSocketProxyService provides access to Apple's DTServiceHub functionality
  * This service enables various instruments and debugging capabilities through the DTX protocol
@@ -37,8 +44,18 @@ export class DVTSecureSocketProxyService extends BaseService {
   private curMessageId: number = 0;
   private readonly channelCache: Map<string, Channel> = new Map();
   private readonly channelMessages: Map<number, ChannelFragmenter> = new Map();
+  private readonly channelWaiters: Map<number, ChannelWaiter[]> = new Map();
   private isHandshakeComplete: boolean = false;
   private readBuffer: Buffer = Buffer.alloc(0);
+  private readonly pendingChunks: Buffer[] = [];
+  private bufferedLength: number = 0;
+  private readonly dataWaiters: Array<() => void> = [];
+  private socketError: Error | null = null;
+  private isPumpRunning: boolean = false;
+  private listeningSocket: net.Socket | null = null;
+  private readonly boundOnData = (chunk: Buffer): void => this.onSocketData(chunk);
+  private readonly boundOnError = (err: Error): void => this.onSocketError(err);
+  private readonly boundOnClose = (): void => this.onSocketClose();
 
   constructor(udid: string) {
     super(udid);
@@ -57,6 +74,7 @@ export class DVTSecureSocketProxyService extends BaseService {
     this.connection = await this.startLockdownWithoutCheckin(DVTSecureSocketProxyService.RSD_SERVICE_NAME);
     this.socket = this.connection.getSocket();
     stripSSL(this.socket);
+    this.attachSocketListeners(this.socket);
 
     await this.performHandshake();
   }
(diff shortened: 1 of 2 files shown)


src/services/ios/dvt/index.ts
```diff
@@ -252,19 +270,104 @@ export class DVTSecureSocketProxyService extends BaseService {
...
+    }
+    this.listeningSocket.off('data', this.boundOnData);
+    this.listeningSocket.off('error', this.boundOnError);
+    this.listeningSocket.off('close', this.boundOnClose);
+    this.listeningSocket = null;
+  }
+
+  private flushDataWaiters(): void {
+    for (const waiter of this.dataWaiters.splice(0)) {
+      waiter();
```
mykola-mokhnach, line 314: can this throw?

src/services/ios/dvt/index.ts
```diff
@@ -252,19 +270,104 @@ export class DVTSecureSocketProxyService extends BaseService {
...
+  private onSocketError(err: Error): void {
+    if (!this.socketError) {
+      this.socketError = err;
+    }
+    this.flushDataWaiters();
+  }
+
+  private onSocketClose(): void {
+    if (!this.socketError) {
+      this.socketError = new Error('Socket closed');
```
mykola-mokhnach, line 345: should we also verify if this was an expected or unexpected close?

src/services/ios/dvt/index.ts
```diff
@@ -252,19 +270,104 @@ export class DVTSecureSocketProxyService extends BaseService {
...
+          waiter.signal.removeEventListener('abort', waiter.onAbort);
+        }
+        waiter.reject(error);
+      }
+    }
+    this.channelWaiters.clear();
+  }
+
+  private hasPendingWaiters(): boolean {
+    for (const waiters of this.channelWaiters.values()) {
```
mykola-mokhnach: this could be simplified using `.some`

src/services/ios/dvt/index.ts
```diff
@@ -354,28 +458,84 @@ export class DVTSecureSocketProxyService extends BaseService {
...
+    signal?.throwIfAborted();
+
+    const promise = new Promise<Buffer>((resolve, reject) => {
+      const waiter: ChannelWaiter = {resolve, reject, signal};
+      if (signal) {
+        const onAbort = () => {
+          const waiters = this.channelWaiters.get(channel);
+          if (waiters) {
+            const idx = waiters.indexOf(waiter);
+            if (idx !== -1) {
```
mykola-mokhnach: nit: I prefer `idx >= 0`

src/services/ios/dvt/index.ts
```diff
@@ -354,28 +458,84 @@ export class DVTSecureSocketProxyService extends BaseService {
...
+        };
+        waiter.onAbort = onAbort;
+        signal.addEventListener('abort', onAbort, {once: true});
       }
 
-      // Check if we have a complete message
-      const message = fragmenter.get();
-      if (message) {
-        return message;
+      let waiters = this.channelWaiters.get(channel);
```
mykola-mokhnach: can it be simplified?

```
this.channelWaiters.set(channel, this.channelWaiters.get(channel) ?? []);
```

src/services/ios/dvt/index.ts
```diff
@@ -354,28 +458,84 @@ export class DVTSecureSocketProxyService extends BaseService {
...
+    }
+    this.isPumpRunning = true;
+    void this.runPump()
+      .catch((err: unknown) => {
+        this.rejectAllWaiters(err instanceof Error ? err : new Error(String(err)));
+      })
+      .finally(() => {
+        this.isPumpRunning = false;
+        if (this.hasPendingWaiters()) {
+          this.ensurePumpRunning();
```
mykola-mokhnach, line 513: this is weird - why this call is recursive?

src/services/ios/dvt/index.ts
harsha509: No. 
Only Promise resolve functions are pushed into `dataWaiters` and calling a `resolver` never throws. 
Any failure in the woken pump surfaces through the `catch` in `ensurePumpRunning`

src/services/ios/dvt/index.ts
harsha509: Expected close never reaches this handler. `close()` detaches the listeners before destroying the socket and rejects pending receivers with its own Service closed error. so this only sees remote or unexpected closes.

If an error caused the close, the error handler has already latched the real error.

src/services/ios/dvt/index.ts
harsha509: fixed

src/services/ios/dvt/index.ts
harsha509: fixed

src/services/ios/dvt/index.ts
harsha509: fixed

src/services/ios/dvt/index.ts
harsha509: It closes a race. 
`runPump` exits when it sees no pending waiters, but a receiver can register between that check and the flag reset here. 
Its `ensurePumpRunning` call no-ops while the flag is still set, so this re-check restarts the pump for it. At most one re-entry per settled pump and it starts a fresh async chain, so there is no stack growth.

Delta456 (approved):

mykola-mokhnach (approved):

PR #301, opened 2026-08-24
feat(accessibility): support writing and resetting device accessibility settings

navin772:
`AccessibilityAuditService` could read the device's nine accessibility settings but not change them. Adds the write side, so an app can be audited under Increase Contrast, Grayscale, Reduce Motion or a larger Dynamic Type without touching the Settings app.

### API

| Method | Notes |
| --- | --- |
| `setAccessibilitySetting(identifier, value, options?)` | `boolean` for a toggle, `0..1` for a slider |
| `resetAccessibilitySettings(options?)` | restores the device's stored defaults |

Also exports `serializeAxSetting` alongside the existing wire serialisers.

⚠️  `resetAccessibilitySettings(options?)` completely resets the user's prior settings if present.

`resetAccessibilitySettings` tests are **opt-in and skipped by default** — the call is persistent and would wipe the accessibility settings of whoever ran the suite:

    ALLOW_ACCESSIBILITY_SETTINGS_RESET=1 UDID=<udid> npm run test:accessibility-audit

### DIFF
--- a/src/services/ios/accessibility-audit/index.ts
+++ b/src/services/ios/accessibility-audit/index.ts
@@ -16,8 +16,10 @@ import {AxAuditDtxTransport, type InvokeOptions} from './dtx-transport.js';
 
 const log = getLogger('AccessibilityAudit');
 
-/** `SettingTypeValue_v1` for a slider setting; every other setting is a toggle (3). */
+/** `SettingTypeValue_v1` for a slider setting, e.g. `DYNAMIC_TYPE`. */
 const SETTING_TYPE_SLIDER = 2;
+/** `SettingTypeValue_v1` for an on/off toggle — every setting bar the slider. */
+const SETTING_TYPE_TOGGLE = 3;
 
 /** `deviceInspectorSetMonitoredEventType:` value that reports focus changes. */
 const MONITORED_EVENT_FOCUS = 2;
@@ -86,7 +88,7 @@ export class AccessibilityAuditService {
     return new AccessibilityAuditService(await AxAuditDtxTransport.connect(udid));
   }
 
-  /** The daemon's API version (26 on iOS 27.0). */
+  /** The daemon's API version (26 on iOS 26.6 and 27.0). */
   async getApiVersion(options?: InvokeOptions): Promise<number> {
     const value = await this.transport.invoke('deviceApiVersion', null, options);
     if (typeof value !== 'number') {
@@ -124,8 +126,10 @@ export class AccessibilityAuditService {
    * service is open — closing it reverts the setting.
    *
    * Sliders snap to the device's tick marks (`DYNAMIC_TYPE` has 12, so `0.5`
-   * reads back as `0.545`). Successive writes to the same setting need a moment
-   * apart; one issued immediately after another is dropped.
+   * reads back as `0.545`). The device acknowledges a write in a few
+   * milliseconds but commits it asynchronously, so back-to-back writes to the
+   * same setting are dropped — ~1.5s apart was reliable in testing, and the
+   * exact minimum is not published.
    *
    * @param identifier A setting identifier, e.g. `INVERT_COLORS`, `DYNAMIC_TYPE`.
    * @param value `boolean` for a toggle, or a number in 0..1 for a slider.
@@ -148,10 +152,16 @@ export class AccessibilityAuditService {
           `Setting "${identifier}" is a slider and expects a number between 0 and 1, got ${JSON.stringify(value)}`,
         );
       }
-    } else if (typeof value !== 'boolean') {
+    } else if (setting.settingType === SETTING_TYPE_TOGGLE) {
       // A toggle takes any truthy value on the wire (0.5 reads back as true),
       // which is too loose to be useful in a typed API.
-      throw new Error(`Setting "${identifier}" is a toggle and expects a boolean, got ${JSON.stringify(value)}`);
+      if (typeof value !== 'boolean') {
+        throw new Error(`Setting "${identifier}" is a toggle and expects a boolean, got ${JSON.stringify(value)}`);
+      }
+    } else {
+      // Only these two types have been observed (iOS 26.6 and 27.0); refuse
+      // rather than guess at the value another type expects.
+      throw new Error(`Setting "${identifier}" has unsupported type ${JSON.stringify(setting.settingType)}`);
     }
 
     const aux = new MessageAux();
@@ -248,7 +258,7 @@ export class AccessibilityAuditService {
    * `hostInspectorCurrentElementChanged:` call, so that is what this reproduces:
    * arm, ask focus to report, wait for the push, disarm. Captured from a live
    * Inspector session — `deviceFetchElementAtNormalizedDeviceCoordinate:`
-   * returns `null` on iOS 27 no matter how it is called.
+   * returns `null` on iOS 26.6 and 27.0 no matter how it is called.
    *
    * @param options Timeout, and whether to draw the on-device highlight.
    */
@@ -346,7 +356,7 @@ export class AccessibilityAuditService {
   /**
    * Returns one of the daemon's well-known elements.
    *
-   * Index `0` and `1` resolve on iOS 27; higher indices return `undefined`.
+   * Index `0` and `1` resolve on iOS 26.6 and 27.0; higher return `undefined`.
    *
    * @param index Which special element to fetch.
    * @param options Reply timeout.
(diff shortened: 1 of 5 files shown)


test/integration/accessibility-audit.spec.ts
```diff
@@ -154,4 +156,230 @@ describe('AccessibilityAuditService', {timeout: 90000}, function () {
...
+    }
+
+    /**
+     * Returns a value only once it has stopped moving.
+     *
+     * A previous run's session override reverts a moment after that service
+     * closes, so a value read immediately can be stale — and writing the value
+     * it is about to revert to produces no observable change.
+     */
+    async function settledValue(identifier: string, timeoutMs = 10000): Promise<unknown> {
```
harsha509: looks like dead code

src/services/ios/accessibility-audit/index.ts
```diff
@@ -116,6 +119,63 @@ export class AccessibilityAuditService {
...
+      // typo would silently max the setting out rather than fail.
+      if (typeof value !== 'number' || !Number.isFinite(value) || value < 0 || value > 1) {
+        throw new Error(
+          `Setting "${identifier}" is a slider and expects a number between 0 and 1, got ${JSON.stringify(value)}`,
+        );
+      }
+    } else if (typeof value !== 'boolean') {
+      // A toggle takes any truthy value on the wire (0.5 reads back as true),
+      // which is too loose to be useful in a typed API.
+      throw new Error(`Setting "${identifier}" is a toggle and expects a boolean, got ${JSON.stringify(value)}`);
```
harsha509: AxDeviceSetting.currentValue is documented as "a boolean toggle, a number, or a string depending on type" 

if setting is settingType it will be misleading

test/integration/accessibility-audit.spec.ts
```diff
@@ -154,4 +156,230 @@ describe('AccessibilityAuditService', {timeout: 90000}, function () {
...
+      assert.ok(ticks > 1, 'the slider should report its tick marks');
+
+      try {
+        // 0.5 is not on a tick, so the device snaps it to the nearest one.
+        const step = 1 / (ticks - 1);
+        const nearest = Math.round(0.5 / step) * step;
+        // 0.5 is not on a tick, so the device snaps it to the nearest one.
+        await service!.setAccessibilitySetting('DYNAMIC_TYPE', 0.5);
+        const settled = (await waitForSetting('DYNAMIC_TYPE', nearest, 10000)) as number;
+        assert.ok(Math.abs(settled - nearest) < 1e-9, `expected ${nearest}, got ${settled}`);
```
harsha509: this asserts the device reports back the exact double you computed (`6/11 = 0.545454…)` within 1e-9. 

If firmware rounds to a displayed decimal (e.g. 0.545), this fails intermittently across OS versions. Safer: assert.ok(Math.abs(settled - nearest) <= step / 2)

src/services/ios/accessibility-audit/index.ts
```diff
@@ -116,6 +119,63 @@ export class AccessibilityAuditService {
     return raw.map(toDeviceSetting);
   }
 
+  /**
+   * Writes one accessibility setting. Applies system-wide, but only while this
+   * service is open — closing it reverts the setting.
+   *
+   * Sliders snap to the device's tick marks (`DYNAMIC_TYPE` has 12, so `0.5`
+   * reads back as `0.545`). Successive writes to the same setting need a moment
```
harsha509: "need a moment apart" ?

callers cant know how long to wait between successive write, we can document the observed minimum gap

test/integration/accessibility-audit.spec.ts
```diff
@@ -154,4 +156,230 @@ describe('AccessibilityAuditService', {timeout: 90000}, function () {
...
+      assert.ok(slider, 'DYNAMIC_TYPE should be present');
+      const original = slider.currentValue as number;
+      const ticks = slider.sliderTickMarks ?? 0;
+      assert.ok(ticks > 1, 'the slider should report its tick marks');
+
+      try {
+        // 0.5 is not on a tick, so the device snaps it to the nearest one.
+        const step = 1 / (ticks - 1);
+        const nearest = Math.round(0.5 / step) * step;
+        // 0.5 is not on a tick, so the device snaps it to the nearest one.
```
harsha509: duplicate comment ?

test/integration/accessibility-audit.spec.ts
```diff
@@ -154,4 +156,230 @@ describe('AccessibilityAuditService', {timeout: 90000}, function () {
...
+   */
+  describe('accessibility settings', function () {
+    /**
+     * Waits for every setting to stop moving before this block starts.
+     *
+     * Overrides written by a previous run revert a moment after that run's
+     * service closes, so back-to-back runs would otherwise read a value that is
+     * still on its way back and assert against the wrong baseline.
+     */
+    before(async function () {
```
harsha509: if outer before fails, this dies'!

test/integration/accessibility-audit.spec.ts
navin772: Yes, a leftover for an old test, will remove

test/integration/accessibility-audit.spec.ts
navin772: will remove one

test/integration/accessibility-audit.spec.ts
navin772: removed it, it was also related to the old test

src/services/ios/accessibility-audit/index.ts
navin772: fixed by isolating check to `SETTING_TYPE_TOGGLE = 3`

test/integration/accessibility-audit.spec.ts
navin772: Updated

src/services/ios/accessibility-audit/index.ts
```diff
@@ -116,6 +121,71 @@ export class AccessibilityAuditService {
...
+   * same setting are dropped — ~1.5s apart was reliable in testing, and the
+   * exact minimum is not published.
+   *
+   * @param identifier A setting identifier, e.g. `INVERT_COLORS`, `DYNAMIC_TYPE`.
+   * @param value `boolean` for a toggle, or a number in 0..1 for a slider.
+   * @param options Reply timeout.
+   * @throws If the identifier is unknown, or the value is wrong for its type.
+   */
+  async setAccessibilitySetting(identifier: string, value: boolean | number, options?: InvokeOptions): Promise<void> {
+    const settings = await this.getAccessibilitySettings(options);
```
harsha509: nit: every write fetches the full settings list just to resolve identifier , can we reduce roundtrip here ?

harsha509 (approved):

src/services/ios/accessibility-audit/index.ts
navin772: good catch, fixed by caching the initial response

src/services/ios/accessibility-audit/index.ts
```diff
@@ -116,6 +129,80 @@ export class AccessibilityAuditService {
...
+   */
+  async setAccessibilitySetting(identifier: string, value: boolean | number, options?: InvokeOptions): Promise<void> {
+    const schema = await this.loadSettingTypes(options);
+    if (!schema.has(identifier)) {
+      throw new Error(
+        `Unknown accessibility setting "${identifier}"; the device supports: ${[...schema.keys()].join(', ')}`,
+      );
+    }
+    const settingType = schema.get(identifier);
+    if (settingType === SETTING_TYPE_SLIDER) {
```
mykola-mokhnach: this seems a bit redundant. Would it be more logical to have a separate class for each setting and then isolate the logic of validating/reading/writing it there?

test/integration/accessibility-audit.spec.ts
```diff
@@ -154,4 +156,187 @@ describe('AccessibilityAuditService', {timeout: 90000}, function () {
...
+   * sequences are not reliable enough to assert, even with retries (retrying
+   * makes it worse, since each retry lands in the previous commit window).
+   * Those behaviours are documented on the methods instead.
+   *
+   * `resetAccessibilitySettings` IS exercised, but only when the device is
+   * already at defaults — see the guard in its own block. It is persistent
+   * rather than session-scoped, so running it against a configured device
+   * would discard real settings.
+   */
+  describe('accessibility settings', function () {
```
mykola-mokhnach, line 177: should we add a skip condition here if the target device os version is below the expected one?

src/services/ios/accessibility-audit/index.ts
navin772: We can isolate this check. But can't create a class for it since the device discovers its catalogue at runtime and from there we get to know what settings exist. Updated.

test/integration/accessibility-audit.spec.ts
navin772: I am not entirely sure (since I don't have an iOS 18 device currently) but I think this service is supported on iOS 18 too, so will leave it like this.

PR #304, opened 2026-08-26
fix(usbmux): handle decoder errors and preserve unconsumed bytes on connect

harsha509:
Fixes three issues in the usbmux layer.

- Malformed frames crashed the process because `parsePlist` errors were not caught. The decoder now emits an error event instead.
- The `LengthBasedSplitter` was piped from the socket but never consumed, so it buffered all traffic. It is removed since the decoder does its own framing.
- `connect()` dropped bytes that arrived after the Result frame in the same chunk. They are now pushed back to the socket so the caller receives them.

Adds unit tests for all three cases.

### DIFF
--- a/src/lib/usbmux/index.ts
+++ b/src/lib/usbmux/index.ts
@@ -6,7 +6,7 @@ import {BaseSocketService} from '../../base-socket-service.js';
 import {getLogger} from '../logger.js';
 import {type PairRecord, processPlistResponse} from '../pair-record/index.js';
 import {type RawPairRecordResponse} from '../pair-record/pair-record.js';
-import {LengthBasedSplitter, parsePlist} from '../plist/index.js';
+import {parsePlist} from '../plist/index.js';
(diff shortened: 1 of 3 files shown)


mykola-mokhnach (approved):

mykola-mokhnach:
do we ever risk buffer overflow? I don't see any limits imposed other to its size

harsha509:
> do we ever risk buffer overflow? I don't see any limits imposed other to its size

HI @mykola-mokhnach ,

it was pre existing, the old cap never protected this path.

> do we ever risk buffer overflow? I don't see any limits imposed other to its size

Hi @mykola-mokhnach, 
this was pre-existing rather than introduced here. 

The old `MAX_FRAME_SIZE` cap never protected the decode path because the splitter output was never consumed, the decoder has always been piped raw from the socket. 

This PR keeps that behavior unchanged and only fixes the crash, the dead splitter, and the byte loss on connect.

PR #305, opened 2026-08-27
fix(port-forwarding): destroy upstream socket when client disconnects during connect

harsha509:
Fixes a socket leak in `DevicePortForwarder`. The `close` and `error` listeners were attached to the local socket only after the upstream connection resolved. A client disconnect during that window was missed, the upstream socket stayed open and both sockets leaked until `stop`. An abrupt client reset in the same window also threw an unhandled `error` event.

Now client `close` and `error` are tracked while the upstream connection is being established and the upstream socket is destroyed immediately if the client is already gone. Includes a unit test covering the early disconnect.

### DIFF
--- a/src/lib/port-forwarding/device-port-forwarder.ts
+++ b/src/lib/port-forwarding/device-port-forwarder.ts
@@ -83,6 +83,21 @@ export class DevicePortForwarder extends EventEmitter {
     this.activeSockets.add(localSocket);
     this.emit('clientConnected', localSocket);
 
+    const upstreamSocket = await this.openUpstreamForClient(localSocket);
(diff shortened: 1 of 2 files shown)


src/lib/port-forwarding/device-port-forwarder.ts
mykola-mokhnach, line 82: this method became too fat. Can we refactor it into smaller logical parts?

mykola-mokhnach (approved):

src/lib/port-forwarding/device-port-forwarder.ts
harsha509: Updated.!

PR #306, opened 2026-08-27
fix(accessibility): reject unknown audit types before they reach the device

navin772:
An audit type the daemon doesn't implement makes it return neither issues nor a completion, and that connection can never run another audit, the service looks alive (`getApiVersion` still works) while every `runAudit` times out.

`runAudit` now validates against `getSupportedAuditTypes()` and throws locally, caching the list per connection. Matching is exact: wrong case, surrounding whitespace and a bare `testType` prefix all wedge the daemon too. An empty list still passes through, the device reads it as "every type".

### DIFF
--- a/src/services/ios/accessibility-audit/index.ts
+++ b/src/services/ios/accessibility-audit/index.ts
@@ -117,6 +117,14 @@ export class AccessibilityAuditService {
    */
   private settingTypes: Map<string, number | undefined> | undefined;
 
+  /**
+   * Supported audit types, read once per connection.
+   *
+   * Like the setting catalogue this is fixed for the device, so validating a
+   * name costs one round trip per connection rather than one per audit.
+   */
+  private auditTypeNames: Set<string> | undefined;
+
   private constructor(private readonly transport: AxAuditDtxTransport) {}
 
   /**
@@ -264,6 +272,7 @@ export class AccessibilityAuditService {
       : undefined;
 
     try {
+      await this.assertKnownAuditTypes(auditTypes, options);
       if (options.targetPid !== undefined) {
         // Narrows the audit to one process; omitted, the daemon uses the
         // foreground app.
@@ -284,6 +293,27 @@ export class AccessibilityAuditService {
       this.auditInFlight = false;
       stopIssues();
       stopLog?.();
+    }
+  }
+
+  /**
+   * Rejects audit types the device does not implement.
+   *
+   * An unrecognised name makes the daemon return neither issues nor a
+   * completion, and that connection can never run another audit — so the name
+   * must never reach the device.
+   */
+  private async assertKnownAuditTypes(auditTypes: string[], options?: InvokeOptions): Promise<void> {
+    if (auditTypes.length === 0) {
+      return;
+    }
+    this.auditTypeNames ??= new Set(await this.getSupportedAuditTypes(options));
+    const unknown = auditTypes.filter((auditType) => !this.auditTypeNames?.has(auditType));
+    if (unknown.length > 0) {
+      throw new Error(
+        `Unknown audit type(s) ${unknown.map((auditType) => JSON.stringify(auditType)).join(', ')}; ` +
+          `the device supports: ${[...(this.auditTypeNames ?? [])].join(', ')}`,
+      );
     }
   }
 
--- a/src/services/ios/accessibility-audit/index.ts
+++ b/src/services/ios/accessibility-audit/index.ts
@@ -272,7 +272,7 @@ export class AccessibilityAuditService {
       : undefined;
 
     try {
-      await this.assertKnownAuditTypes(auditTypes, options);
+      await this.assertKnownAuditTypes(auditTypes);
       if (options.targetPid !== undefined) {
         // Narrows the audit to one process; omitted, the daemon uses the
         // foreground app.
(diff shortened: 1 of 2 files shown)


src/services/ios/accessibility-audit/index.ts
```diff
@@ -287,6 +296,27 @@ export class AccessibilityAuditService {
...
+   * must never reach the device.
+   */
+  private async assertKnownAuditTypes(auditTypes: string[], options?: InvokeOptions): Promise<void> {
+    if (auditTypes.length === 0) {
+      return;
+    }
+    this.auditTypeNames ??= new Set(await this.getSupportedAuditTypes(options));
+    const unknown = auditTypes.filter((auditType) => !this.auditTypeNames?.has(auditType));
+    if (unknown.length > 0) {
+      throw new Error(
```
mykola-mokhnach, line 324: can it ever happen that this.auditTypeNames is empty?

src/services/ios/accessibility-audit/index.ts
```diff
@@ -287,6 +296,27 @@ export class AccessibilityAuditService {
...
+        `Unknown audit type(s) ${unknown.map((auditType) => JSON.stringify(auditType)).join(', ')}; ` +
```
mykola-mokhnach: consider using util.pluralize

mykola-mokhnach (approved):

src/services/ios/accessibility-audit/index.ts
```diff
@@ -264,6 +272,7 @@ export class AccessibilityAuditService {
       : undefined;
 
     try {
+      await this.assertKnownAuditTypes(auditTypes, options);
```
harsha509: dont forward options here...

the caller's audit-completion `timeoutMs` becomes the reply timeout for the `deviceAllSupportedAuditTypes` invoke (overriding the 15s transport default), so a long audit budget can block minutes in pre-flight and total wait approaches 2× timeoutMs.

test/integration/accessibility-audit.spec.ts
```diff
@@ -217,20 +281,29 @@ describe('AccessibilityAuditService', {timeout: 300000}, function () {
       }
     });
 
-    it('rejects an unknown setting identifier instead of silently doing nothing', async function () {
+    it('rejects an unknown setting identifier instead of silently doing nothing', async function (t) {
+      if (skipWithoutSettingWrites(t)) {
```
harsha509: These three tests reject client side before the write selector is ever invoked, so they shouldnt be skipped behind `skipWithoutSettingWrites`

test/integration/accessibility-audit.spec.ts
```diff
@@ -269,6 +342,9 @@ describe('AccessibilityAuditService', {timeout: 300000}, function () {
 
       /** Reports why a test is being skipped, or null when it may run. */
       async function skipReason(): Promise<string | null> {
+        if (!supportsSettingWrites) {
```
harsha509: this gates on `deviceUpdateAccessibilitySetting`
but 2-3 tests only use `deviceUpdateAccessibilitySetting` gates on reset selector.

may be add the write only for clear session override selector

test/integration/accessibility-audit.spec.ts
```diff
@@ -174,7 +183,59 @@ describe('AccessibilityAuditService', {timeout: 300000}, function () {
...
+    const POISON: Array<[string, string[]]> = [
+      ['an invented name', ['bogusAuditTypeThatDoesNotExist']],
+      ['the wrong case', ['TESTTYPECONTRAST']],
+      ['an empty string', ['']],
+      ['a bare prefix', ['testType']],
+      ['surrounding whitespace', [' testTypeContrast ']],
+    ];
+
+    for (const [label, auditTypes] of POISON) {
+      it(`rejects ${label} without contacting the device`, async function () {
```
harsha509: on a cold cacge the first rejection doesnt contact the device here

test/integration/accessibility-audit.spec.ts
```diff
@@ -174,7 +183,59 @@ describe('AccessibilityAuditService', {timeout: 300000}, function () {
...
+      });
+    }
+
+    it('rejects a valid type mixed with an unknown one', async function () {
+      const types = await service!.getSupportedAuditTypes();
+
+      await assert.rejects(() => service!.runAudit([types[0], 'bogusType'], {timeoutMs: 20000}), /Unknown audit type/);
+    });
+
+    it('still accepts an empty list, which the device reads as every type', async function () {
```
harsha509: this duplicates above test in file "accepts an empty audit-type list without hanging"

test/integration/accessibility-audit.spec.ts
```diff
@@ -174,7 +183,59 @@ describe('AccessibilityAuditService', {timeout: 300000}, function () {
...
+      // Rejecting this would be a regression: the daemon audits everything.
+      assert.ok(Array.isArray(await service!.runAudit([], {timeoutMs: 60000})));
+    });
+
+    it('leaves the connection usable after a rejected audit', async function (t) {
+      const types = await service!.getSupportedAuditTypes();
```
harsha509: The recovery audit only needs to prove the connection still works. the purpose of test is only checks that the connection still works after a rejection, checkignf for all is a time waste 

may be

```
it('leaves the connection usable after a rejected audit', async function () {
  await assert.rejects(
    service!.runAudit(['nopeNotAType'], {timeoutMs: 20000}),
    /Unknown audit type/,
  );
  const issues = await service!.runAudit(['testTypeContrast'], {timeoutMs: 60000});
  assert.ok(Array.isArray(issues));
});
```

harsha509 (approved):

test/integration/accessibility-audit.spec.ts
navin772: will rename test accordingly

test/integration/accessibility-audit.spec.ts
navin772: will remove

src/services/ios/accessibility-audit/index.ts
navin772: Shouldn't be, but it isn't guaranteed, and an empty list would reject valid types and stay cached. Will update so an empty result is neither cached nor validated against.

PR #308, opened 2026-08-29
fix(deps): require appium-ios-tuntap ^2.0.0

harsha509:
tuntap 2.0.0 removed the unused JS polling API. remotexpc uses only the tunnel APIs, so no code changes are needed.


harsha509:
only version bump.
Should be good

LGTM!

PR #310, opened 2026-08-31
feat(accessibility-audit): add focus traversal via `moveFocus` and `walkElements`

navin772:
Adds focus traversal to the accessibility audit service: `moveFocus(direction)` moves VoiceOver-style focus (First/Last/Next/Previous) and returns the element it lands on, and `walkElements()` enumerates a screen by walking focus forward.

Focus cycles rather than reporting an end, and nothing the daemon sends identifies an element - `platformElement` is a fresh handle per focus event and `accessibilityIdentifier` is absent on most elements. The walk therefore stops when a pair of announcements recurs, collapsing runs of identical announcements (a photo grid announces "Photo, Image" repeatedly) and holding one element back so the revisit isn't yielded.

### API Usage

```ts
const audit = await Services.startAccessibilityAuditService(udid);
try {
  // One step — move focus and see what it landed on.
  const focused = await audit.moveFocus(AxFocusDirection.Next);
  console.log(focused.caption);        // 'Wi-Fi, **SSID name**, Button'

  // Whole screen — walk focus forward until the sequence repeats.
  for await (const element of audit.walkElements()) {
    console.log(element.caption);
  }
} finally {
  audit.close();
}
```

### DIFF
--- a/src/services/ios/accessibility-audit/index.ts
+++ b/src/services/ios/accessibility-audit/index.ts
@@ -62,15 +62,12 @@ const MONITORED_EVENT_OFF = 0;
  * Where {@link AccessibilityAuditService.moveFocus} sends the accessibility
  * focus. These are the daemon's own `direction` values, verified live.
  */
-export const AxFocusDirection = {
-  Previous: 3,
-  Next: 4,
-  First: 5,
-  Last: 6,
-} as const;
-
-/** A value of {@link AxFocusDirection}. */
-export type AxFocusDirection = (typeof AxFocusDirection)[keyof typeof AxFocusDirection];
+export enum AxFocusDirection {
+  Previous = 3,
+  Next = 4,
+  First = 5,
+  Last = 6,
+}
 
 /** Safety valve for {@link AccessibilityAuditService.walkElements}. */
 const MAX_WALK_ELEMENTS = 1000;
--- a/src/services/ios/accessibility-audit/index.ts
+++ b/src/services/ios/accessibility-audit/index.ts
@@ -72,6 +72,12 @@ export enum AxFocusDirection {
 /** Default wait for the focus event a move produces. */
 const DEFAULT_FOCUS_TIMEOUT_MS = 15000;
 
+/**
+ * How many further elements must keep matching before a repeated step is taken
+ * as proof the walk has come round again.
+ */
+const WALK_PERIOD_CONFIRM_STEPS = 6;
+
 /** Safety valve for {@link AccessibilityAuditService.walkElements}. */
 const MAX_WALK_ELEMENTS = 1000;
 
@@ -538,9 +544,15 @@ export class AccessibilityAuditService {
    * therefore collapsed for repeat detection and still yielded individually,
    * and the walk stops only when a *pair* of announcements recurs.
    *
-   * The consequence worth knowing: a screen that legitimately repeats the same
-   * two announcements in sequence ends the walk early, and one where every
-   * element announces alike runs to the safety limit.
+   * A repeated step is treated as a hypothesis rather than proof: the walk keeps
+   * going and only stops once the announcements have kept matching one period
+   * back for several more elements. A screen that happens to take the same step
+   * twice is therefore walked in full, where believing the first repeat would
+   * have cut it short.
+   *
+   * The consequence worth knowing: a screen whose announcements repeat over a
+   * long enough stretch to satisfy that check still ends early, and one where
+   * every element announces alike runs to the safety limit.
    *
    * A screen with nothing focusable yields nothing. Note the daemon reports what
    * is actually drawn, so an app that has not rendered looks empty.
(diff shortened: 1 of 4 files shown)


src/services/ios/accessibility-audit/index.ts
```diff
@@ -410,6 +435,177 @@ export class AccessibilityAuditService {
...
+  async *walkElements(options: InspectOptions = {}): AsyncGenerator<AxInspectedElement, void, unknown> {
+    const steps = new Set<string>();
+    let previous: string | undefined;
+    let pending: AxInspectedElement | undefined;
+    let direction: AxFocusDirection = AxFocusDirection.First;
+
+    for (let step = 0; step < MAX_WALK_ELEMENTS; step += 1) {
+      let element: AxInspectedElement;
+      try {
+        element = await this.moveFocus(
```
harsha509: A screen with one focusable element rejects instead of yielding it.

src/services/ios/accessibility-audit/index.ts
```diff
@@ -410,6 +435,177 @@ export class AccessibilityAuditService {
...
+          direction,
+          step === 0
+            ? {
+                ...options,
+                timeoutMs: Math.min(options.timeoutMs ?? FOCUS_FIRST_PROBE_TIMEOUT_MS, FOCUS_FIRST_PROBE_TIMEOUT_MS),
```
harsha509: A single transient timeout mid-walk discards the element already in hand.

src/services/ios/accessibility-audit/index.ts
```diff
@@ -410,6 +435,177 @@ export class AccessibilityAuditService {
...
+          yield pending;
+        }
+        return;
+      }
+
+      if (previous === undefined) {
+        previous = announcement;
+      } else if (announcement !== previous) {
+        const transition = `${previous}\u0000${announcement}`;
+        if (steps.has(transition)) {
```
harsha509: The JSDoc notes that `a screen that legitimately repeats the same two announcements in sequence ends the walk early`. Measured on a 5-element screen announcing 

```
A, B, A, B, C:


yielded [A, B] — 2 of 5

File: src/services/ios/accessibility-audit/index.ts [hunk 4]
```diff
@@ -410,6 +435,177 @@ export class AccessibilityAuditService {
...
+   * push back to the move that caused it. A push landing after the wait gave up
+   * is delivered to the next caller, which then sees a stale element — so treat
+   * a timeout as "this connection is briefly out of step" and let it settle.
+   * Draining blindly does not work: a move that changes nothing is answered with
+   * silence, so the event discarded would as often be a real one.
+   *
+   * @param direction Where to move; see {@link AxFocusDirection}.
+   * @param options Timeout, and whether to draw the on-device highlight.
+   */
+  async moveFocus(direction: AxFocusDirection, options: InspectOptions = {}): Promise<AxInspectedElement> {
```
harsha509, line 471: moveFocus has no in-flight guard, unlike runAudit.

src/services/ios/accessibility-audit/index.ts
harsha509: A slow opening move makes the walk duplicate and reorder elements, and it breaks this PR's own test invariant.

If the First move is answered after the 2000ms probe gives up, its push lands during a later step's wait and every subsequent step reads one push behind. Same 6-element screen, only the opening move's latency changed (300ms per move otherwise):


opening move answered in  100ms → yielded [A, B, C, D, E, F]
opening move answered in 1000ms → yielded [B, A, C, D, E, F, A, B]

src/services/ios/accessibility-audit/index.ts
```diff
@@ -410,6 +435,177 @@ export class AccessibilityAuditService {
...
+   *   // Read now: after the next iteration this element is no longer focused.
+   *   const value = label && focused.element
+   *     ? await audit.getElementAttributeValue(focused.element, label)
+   *     : undefined;
+   * }
+   * ```
+   *
+   * @param options Timeout per step, and whether to draw the on-device highlight.
+   */
+  async *walkElements(options: InspectOptions = {}): AsyncGenerator<AxInspectedElement, void, unknown> {
```
harsha509, line 581: Monitoring is armed and disarmed on every step of the walk.

Because walkElements goes through `moveFocus`, each step sends `deviceInspectorSetMonitoredEventType:`  twice. Counted on a 3-element walk:


`15 oneway frames total; deviceInspectorSetMonitoredEventType: sent 10x [2,0,2,0,2,0,2,0,2,0]`

src/services/ios/accessibility-audit/index.ts
```diff
@@ -410,6 +435,177 @@ export class AccessibilityAuditService {
...
+    const steps = new Set<string>();
+    let previous: string | undefined;
+    let pending: AxInspectedElement | undefined;
+    let direction: AxFocusDirection = AxFocusDirection.First;
+
+    for (let step = 0; step < MAX_WALK_ELEMENTS; step += 1) {
```
harsha509: A screen with one focusable element rejects instead of yielding it.

src/services/ios/accessibility-audit/index.ts
navin772: Fixed

src/services/ios/accessibility-audit/index.ts
navin772: added a guard similar to `runAudit`

src/services/ios/accessibility-audit/index.ts
navin772: Fixed

src/services/ios/accessibility-audit/index.ts
navin772: `deviceInspectorMoveWithOptions:` is a oneway call. The answer doesn't come back as a reply to that call - it arrives as a separate, unsolicited inbound call, `hostInspectorCurrentElementChanged:`, carrying no request id, no sequence number, nothing linking it to the move that triggered it.

Updated the docs as per this

src/services/ios/accessibility-audit/index.ts
navin772: came across this, trying to find a good solution

navin772: Fixed. A repeated step is now treated as a candidate period rather than proof of a wrap: the walk keeps going and only stops once the announcements have kept matching one period back for 6 more elements.

harsha509 (approved):

PR #312, opened 2026-09-01
fix(remote-xpc): propagate XPC decode errors instead of stalling

harsha509:
### Problem

A single undecodable XPC message stalled the connection. `ingestXpcData` caught
every decode failure and buffered the bytes as if the message were incomplete, so
the parser waited forever for bytes that were already there and every later message
on that stream was lost. All streams shared one buffer, so interleaved frames could
also corrupt each other.

Decode failures were also reported on `'error'`. Node throws an `'error'` with no
listener, and `RsdServiceCatalogClient` used `once('error')`, so a later error
crashed the process. `CoreDeviceService` attached its listener after `connect()`,
leaving the handshake window uncovered.

### Fix

- Add `probeXpcFraming`, a non throwing framing check that separates "wait for more
  bytes" from "stream is unusable". A body length above the initial receive window
  is treated as desync, not a partial read.
- Buffer pending bytes per HTTP/2 stream.
- Emit `'decodeError'` for one unreadable message and keep the connection alive.
  Reserve `'error'` for desync, which latches the transport closed.
- Ignore socket events from a socket already replaced by a reconnect.
- Attach the permanent listeners before `connect()`, idempotently.
- Surface the last decode error in the CoreDevice invoke timeout message.

### Tests

New unit coverage for the framed transport, the RSD catalog client, the service
catalog collector and CoreDeviceService. 836 unit tests pass.

### DIFF
--- a/src/lib/remote-xpc/remote-xpc-framed-transport.ts
+++ b/src/lib/remote-xpc/remote-xpc-framed-transport.ts
@@ -237,7 +237,9 @@ export class RemoteXpcFramedTransport extends EventEmitter {
   private handleDesync(reason: string): void {
     this.desynced = true;
     this.connected = false;
-    void this.close();
+    this.close().catch((error: unknown) => {
+      log.debug(`Failed to close a desynced RemoteXPC transport: ${error}`);
+    });
     this.emit('error', new Error(reason));
   }
 
--- a/src/lib/remote-xpc/remote-xpc-framed-transport.ts
+++ b/src/lib/remote-xpc/remote-xpc-framed-transport.ts
@@ -2,10 +2,11 @@ import {EventEmitter} from 'node:events';
 import net from 'node:net';
 
 import {getLogger} from '../logger.js';
+import type {XPCDictionary} from '../types.js';
 import {DataFrame} from './handshake-frames.js';
 import Handshake from './handshake.js';
 import {Http2FrameParser, buildWindowUpdateFrames} from './http2-frame-parser.js';
-import {decodeMessage} from './xpc-protocol.js';
+import {decodeMessage, probeXpcFraming} from './xpc-protocol.js';
 
 const log = getLogger('RemoteXpcFramedTransport');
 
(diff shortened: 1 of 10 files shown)


src/lib/remote-xpc/remote-xpc-framed-transport.ts
```diff
@@ -168,29 +185,77 @@ export class RemoteXpcFramedTransport extends EventEmitter {
...
+
+  /**
+   * Latches the whole connection closed: framing alignment is lost and the peer's
+   * remaining bytes cannot be trusted on any stream. Teardown starts before the
+   * emit so a listener that reconnects synchronously keeps its new socket.
+   */
+  private handleDesync(reason: string): void {
+    this.desynced = true;
+    this.connected = false;
+    void this.close();
```
mykola-mokhnach: is close safe to call async?

src/lib/remote-xpc/service-catalog.ts
```diff
@@ -16,39 +16,55 @@ export interface ServicesResponse {
...
+      const message = pending.subarray(0, framing.byteLength);
+      pending = pending.subarray(framing.byteLength);
 
-        const catalog = servicesFromXpcBody(message.body);
-        if (catalog) {
-          this.previousFrameData = pending;
-          return catalog;
-        }
-      } catch {
+      const catalog = extractCatalog(message);
```
mykola-mokhnach, line 48: we should be careful here to not put too much logging on the hot path

mykola-mokhnach (approved):

src/lib/remote-xpc/remote-xpc-framed-transport.ts
harsha509: Fixed. Added a `.catch()`.

src/lib/remote-xpc/service-catalog.ts
harsha509: Fixed. Switched the decode error logs to `once`.

PR #315, opened 2026-09-01
fix(dvt): correlate DTX replies so device callbacks don't desync request/reply

navin772:
## Problem
A DTX channel carries replies and device-initiated callbacks over the same stream. The pump queued every message for a channel into one FIFO and `receivePlist()` popped the head, with no correlation between a request and its reply - `conversationIndex` was parsed in the header and never used. One stray callback shifted every later read by one message, permanently.

## Change
Opt-in reply correlation. `receivePlist()` is deliberately unchanged - sysmontap, network-monitor, graphics, energy-monitor and notifications legitimately read `conversationIndex === 0` messages.

- `ChannelFragmenter` queues `{header, data}`; `get(filter?)` returns the first *matching* message and leaves earlier ones queued.
- Waiters carry an optional filter; `deliverQueuedMessages()` leaves a waiter pending when nothing matches, so the pump keeps reading. Skipped callbacks stay queued for `ProcessControl.outputEvents()`.
- New `recvReplyPlist()` / `Channel.receiveReply()`. `recvReplyPlist` is optional on `DTXServiceProvider`, so testmanagerd's provider falls back.
- ProcessControl's four request/reply sites use `receiveReply()`.

Also fixed alongside:

- **process-control.spec**: the Settings test asserted `pid > 0` without launching Settings; look-ups report 0 for an app that isn't running.
- **device-info.spec**: `ctx.skip()` was followed by an unconditional `throw`, so the test skipped *and* failed. `nameForGID:` is genuinely gone from `DTDeviceInfoService` after iOS 14.x, so skipping is correct.


src/services/ios/dvt/channel.ts
```diff
@@ -37,6 +39,18 @@ export class Channel {
...
   }
 
+  /**
+   * Receive the reply to a request sent on this channel, skipping callbacks the
+   * device sent on its own initiative. Prefer this over {@link receivePlist}
+   * for request/reply, which returns whichever message arrived first.
+   * @param signal Optional AbortSignal for cancellation
+   */
+  async receiveReply(signal?: AbortSignal): Promise<any> {
+    const [data] = await (this.service.recvReplyPlist?.(this.channelCode, signal) ??
```
mykola-mokhnach: are these methods guaranteed to always return a list?

src/services/ios/dvt/index.ts
```diff
@@ -552,19 +582,33 @@ export class DVTSecureSocketProxyService extends BaseService {
...
+   * nothing stays pending so the pump keeps reading.
+   */
+  private deliverQueuedMessages(channel: number): void {
+    const waiters = this.channelWaiters.get(channel);
+    const fragmenter = this.channelMessages.get(channel);
+    if (!waiters?.length || !fragmenter) {
+      return;
+    }
+
+    for (let i = 0; i < waiters.length;) {
```
mykola-mokhnach: nit: I would prefer a while loop

test/integration/dvt-instruments/device-info.spec.ts
```diff
@@ -153,8 +153,10 @@ describe('DeviceInfo Instrument', {timeout: 30000}, function () {
         assert.ok(groupName.length > 0);
       } catch (error) {
         const message = (error as Error).message;
+        // Apple dropped nameForGID: from DTDeviceInfoService
         if (message.includes('nameForGID') && message.includes('does not respond')) {
           ctx.skip();
+          return;
```
mykola-mokhnach: afaik we can combine return and ctx.skip()

mykola-mokhnach (approved):

src/services/ios/dvt/channel.ts
navin772: yes

src/services/ios/dvt/index.ts
```diff
@@ -552,19 +582,34 @@ export class DVTSecureSocketProxyService extends BaseService {
...
-        }
+      this.deliverQueuedMessages(receivedChannel);
+    }
+  }
+
+  /**
+   * Hand queued messages to the waiters that accept them. A waiter matching
+   * nothing stays pending so the pump keeps reading.
+   */
+  private deliverQueuedMessages(channel: number): void {
```
harsha509, line 599: the reply filter is not correlated to a request, so an unfiltered waiter already pending on the same channel takes the reply and `receiveReply()` then never resolves.

src/services/ios/dvt/index.ts
```diff
@@ -552,19 +582,34 @@ export class DVTSecureSocketProxyService extends BaseService {
...
+    const waiters = this.channelWaiters.get(channel);
+    const fragmenter = this.channelMessages.get(channel);
+    if (!waiters?.length || !fragmenter) {
+      return;
+    }
+
+    let i = 0;
+    while (i < waiters.length) {
+      const waiter = waiters[i];
+      const message = fragmenter.get(waiter.filter);
```
harsha509: This picks by filter but not by request. A waiter with no filter matches anything, including a reply, so whoever is first in `waiters` wins.

`ProcessControl.outputEvents()` sits in `receivePlistWithAux()` (unfiltered) on the same channel as the four `receiveReply()` sites. If it is pending when a reply lands, it takes the reply and the `receiveReply()` waiter is left with nothing that can ever match.

Same scenario, this branch vs merge base. Pending unfiltered `recvMessage(1)`, then a filtered one, then a frame with `conversationIndex=1`, then a later frame with `conversationIndex=0`:

    base:  stream=12  replyResolved=true   replyLen=99
    here:  stream=12  replyResolved=false  replyLen=undefined

On base the caller got a wrong value. Here it never returns, because no `conversationIndex=0` frame can pass the filter. None of `launch()`, `getPidForBundleIdentifier()`, `signal()`,`disableMemoryLimitForPid()` passes a signal or timeout, so it is an unbounded hang.

Fix: correlate on `header.identifier`, not on `conversationIndex > 0`. `src/services/ios/accessibility-audit/dtx-transport.ts` already keys pending replies by identifier and cannot hit this.

src/services/ios/dvt/index.ts
```diff
@@ -188,7 +193,31 @@ export class DVTSecureSocketProxyService extends BaseService {
...
     signal?: AbortSignal,
   ): Promise<[any, any[]]> {
-    const [data, aux] = await this.recvMessage(channel, signal);
+    return await this.recvPlistFiltered(channel, signal);
+  }
+
+  /**
+   * Receive the reply to a request, skipping device-initiated callbacks. One
+   * stray callback would otherwise shift every later read by one message.
+   * Skipped callbacks stay queued for stream readers like
```
harsha509: This says skipped callbacks stay queued for readers like `outputEvents()`, which reads as "the two coexist". They do not. See the comment on line 603.

src/services/ios/dvt/channel.ts
```diff
@@ -37,6 +39,19 @@ export class Channel {
     return data;
   }
 
+  /**
+   * Receive the reply to a request sent on this channel, skipping callbacks the
+   * device sent on its own initiative. Prefer this over {@link receivePlist}
+   * for request/reply, which returns whichever message arrived first.
+   * @param signal Optional AbortSignal for cancellation
+   */
+  async receiveReply(signal?: AbortSignal): Promise<any> {
```
harsha509: Silent fallback to the uncorrelated `recvPlist` when a provider omits `recvReplyPlist`. `DvtTestmanagedProxyService` is one, so a caller there gets old behaviour with no signal that correlation is off.

harsha509 (changes requested):

src/services/ios/dvt/index.ts
navin772: Fixed, also added unit tests for this

Delta456 (approved):

harsha509 (approved):

PR #316, opened 2026-09-01
fix(remote-xpc): honor HTTP/2 frame size and flow control on outbound DATA frames

harsha509:
## Problem

`sendDataFrame` wrote every XPC message as one DATA frame with no size limit and no window accounting. The frame parser dropped the peer's SETTINGS and WINDOW_UPDATE frames.

## Observed on device

iPhone on iOS 26.6.1. A script invoked `com.apple.coredevice.feature.getdeviceinfo` with a random blob of the given size in the request input, so the device had to accept every DATA frame before it could reply.

| Request padding | Before | After |
|---|---|---|
| 0 bytes | reply, 11 attributes | reply, 11 attributes |
| 20,000 bytes | read ECONNRESET | reply |
| 300,000 bytes | write EPIPE | reply |
| 3,000,000 bytes | not run | reply in 154ms |

The 20,000 byte case is under the 65535 window but over the 16384 frame limit, so frame size alone triggers the reset.

## Fix

- Split outbound payloads at the peer's max frame size, default 16384
- Track the connection and per-stream send windows, default 65535, and hold the remainder until a WINDOW_UPDATE arrives
- Parse peer SETTINGS for MAX_FRAME_SIZE and INITIAL_WINDOW_SIZE
- Debit the windows for the DATA frames the handshake already sent

3 new unit tests against a loopback peer fail before and pass after. `test:coredevice-device-info` and `test:app-service` pass on device.


src/lib/remote-xpc/http2-frame-parser.ts
```diff
@@ -67,6 +81,15 @@ function parseFrame(buffer: Buffer): ParsedFrame {
   };
 }
 
+/** SETTINGS body is a list of 16-bit identifier / 32-bit value pairs (RFC 7540 §6.5.1). */
+function parseSettings(body: Buffer): Record<number, number> {
+  const settings: Record<number, number> = {};
```
mykola-mokhnach: this can be simplified using reduce

src/lib/remote-xpc/handshake.ts
```diff
@@ -11,13 +11,19 @@ export type MessageId = number;
 class Handshake {
   private _socket: Socket;
   private readonly _nextMessageId: Record<ChannelId, MessageId>;
+  /** DATA payload bytes sent per stream, so the transport can debit its send windows. */
+  readonly dataBytesSent: Record<ChannelId, number>;
```
mykola-mokhnach: I am not sure about this field being public. It breaks incapsulation

src/lib/remote-xpc/remote-xpc-framed-transport.ts
```diff
@@ -88,7 +98,71 @@ export class RemoteXpcFramedTransport extends EventEmitter {
...
+  }
+
+  /**
+   * Writes queued payloads as DATA frames no larger than the peer's max frame size
+   * or its send windows. Stops at the first blocked payload so ordering holds.
+   */
+  private flushPendingSends(): void {
+    const socket = this.socket;
+    while (socket?.writable && this.pendingSends.length > 0) {
+      const send = this.pendingSends[0];
```
mykola-mokhnach: is it a FIFO buffer?

test/unit/remote-xpc/remote-xpc-framed-transport.spec.ts
```diff
@@ -97,6 +99,74 @@ async function withConnectedTransport(
   }
 }
 
+/** RFC 7540 defaults the peer has not overridden: SETTINGS_MAX_FRAME_SIZE and the initial flow-control window. */
+const DEFAULT_MAX_FRAME_SIZE = 16384;
```
mykola-mokhnach: why do we repeat these values here instead of importing them from constants?

src/lib/remote-xpc/http2-frame-parser.ts
harsha509: HI @mykola-mokhnach ,

Kept the loop because here input is `buffer`, walked in 6-byte stripes and not an array so...

test/unit/remote-xpc/remote-xpc-framed-transport.spec.ts
harsha509: removed and imported from Http2Constants

src/lib/remote-xpc/remote-xpc-framed-transport.ts
harsha509: Yes Mykola,

`sendDataFrame `pushes to the back, `flushPendingSends` always takes index 0 and stops at the first blocked payload, so ordering is preserved.

also renamed to `sendQueue` to amke sure intent is visible

src/lib/remote-xpc/handshake.ts
harsha509: fixed it

src/lib/remote-xpc/handshake.ts
```diff
@@ -11,13 +11,23 @@ export type MessageId = number;
...
       [Http2Constants.REPLY_CHANNEL]: 0,
     };
+    this._dataBytesSent = {
+      [Http2Constants.ROOT_CHANNEL]: 0,
+      [Http2Constants.REPLY_CHANNEL]: 0,
+    };
+  }
+
+  /** DATA payload bytes sent per stream, so the transport can debit its send windows. */
+  get dataBytesSent(): Readonly<Record<ChannelId, number>> {
```
mykola-mokhnach, line 29: this getter still allows mutation of internal property, would it make sense to make a copy instead?

src/lib/remote-xpc/http2-frame-parser.ts
mykola-mokhnach: how about:

```js
function parseSettings(body: Buffer): Record<number, number> {
  return [...Array(Math.floor(body.length / SETTINGS_ENTRY_SIZE))]
    .reduce<Record<number, number>>((settings, _, i) => {
      const offset = i * SETTINGS_ENTRY_SIZE;
      settings[body.readUInt16BE(offset)] = body.readUInt32BE(offset + 2);
      return settings;
    }, {});
}
```

mykola-mokhnach (approved):

src/lib/remote-xpc/handshake.ts
harsha509: Makes sense, Updated

src/lib/remote-xpc/http2-frame-parser.ts
harsha509: mm works Mykola.
updated with same

PR #317, opened 2026-09-02
fix(remote-xpc): reassemble inbound XPC messages in linear time

harsha509:
Inbound XPC reassembly copied the whole pending buffer twice per 16 KiB DATA frame, so cost grew with the square of the message size.

Changes
- Keep a chunk list and running length per stream, join once when the declared length is reached
- Skip empty DATA frames, copy retained chunks so a stalled stream cannot pin the parser buffer
- probeXpcFraming returns byteLength on incomplete input once the header is readable
- Tests for copy volume, empty frame handling and retained chunk ownership

Loopback through the real socket path, 16 KiB frames, time blocked in reassembly
| message | before | after |
|---|---|---|
| 4 MiB | 19.5 ms | 0.7 ms |
| 8 MiB | 234.8 ms | 1.4 ms |
| 15 MiB | 933.2 ms | 2.9 ms |

Unit suite 842 pass. Device suites app-service, coredevice-device-info and tunnel pass on both builds.

### DIFF
--- a/src/lib/remote-xpc/remote-xpc-framed-transport.ts
+++ b/src/lib/remote-xpc/remote-xpc-framed-transport.ts
@@ -7,7 +7,7 @@ import {Http2Constants} from './constants.js';
 import {DataFrame} from './handshake-frames.js';
 import Handshake from './handshake.js';
 import {Http2FrameParser, buildWindowUpdateFrames} from './http2-frame-parser.js';
-import {decodeMessage, probeXpcFraming} from './xpc-protocol.js';
+import {decodeMessage, probeXpcFraming, XPC_WRAPPER_HEADER_SIZE} from './xpc-protocol.js';
 
 const log = getLogger('RemoteXpcFramedTransport');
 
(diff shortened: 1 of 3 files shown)


src/lib/remote-xpc/remote-xpc-framed-transport.ts
```diff
@@ -283,33 +290,63 @@ export class RemoteXpcFramedTransport extends EventEmitter {
...
-      this.decodeAndEmit(message);
+      if (framing.status === 'incomplete') {
+        this.pendingXpcData.set(streamId, {
+          chunks: [Buffer.from(rest)],
+          length: rest.length,
+          byteLength: framing.byteLength,
+        });
+        return;
+      }
+      this.decodeAndEmit(rest.subarray(0, framing.byteLength));
```
mykola-mokhnach, line 348: should we also verify that `framing.byteLength > 0` to avoid infinite loop?

mykola-mokhnach (approved):

src/lib/remote-xpc/remote-xpc-framed-transport.ts
harsha509: Good questn,

But this cannot happen Mykola. `byteLength` is the `24 byte` header plus an unsigned body length, so an `ok probe` result is always at least `24` and the loop always moves forward. 

Adding a check here would guard a value the probe cannot produce

PR #318, opened 2026-09-03
fix(dvt): correlate remaining instrument acks with `receiveReply()`

navin772:
Follow-up to #315, which added identifier-based reply correlation but only migrated `ProcessControl`. The remaining instruments still read their acks through unfiltered `receivePlist()`, which takes whichever message arrives first, so a callback, or another in-flight request's reply, can be picked up instead.

Migrates the 10 request/reply sites now that `Channel.call()` returns the message identifier:

- `condition-inducer` (3), `activity-trace-tap` (2), `graphics`, `screenshot`, `application-listing`, `energy-monitor`, `location-simulation` and `device-info`
 
Stream readers keep using unfiltered reads and must not be migrated: `graphics.messages()`, `sysmontap`, `network-monitor`, `notifications`, `process-control.outputEvents()`.

These channels aren't broken today and nothing streams on them concurrently, so this is closing a latent bug, not a live one.

### DIFF
--- a/src/services/ios/dvt/instruments/activity-trace-tap.ts
+++ b/src/services/ios/dvt/instruments/activity-trace-tap.ts
@@ -108,10 +108,10 @@ export class ActivityTraceTap extends BaseInstrument {
       ur: 500,
     };
     const channel = this.requireChannel();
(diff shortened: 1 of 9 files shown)


harsha509 (approved):
Looks good to me! 👍

PR #319, opened 2026-09-03
fix(mobile-image-mounter): handle connection errors and cleanup service after tests

navin772:


src/services/ios/mobile-image-mounter/index.ts
```diff
@@ -330,11 +330,16 @@ class MobileImageMounterService extends BaseService implements MobileImageMounte
...
 
     if (isNewConnection && res?.Request === 'StartService') {
-      return await conn.receive();
+      res = await conn.receive();
     }
+
+    if (res?.Error) {
+      this.closeConnection();
+    }
+
```
harsha509: may be we can rewrite as

```
private async sendRequest(request: PlistDictionary, timeout?: number): Promise<PlistDictionary> {
  try {
    const res = await this.exchange(request, timeout);
    if (res?.Error) {
      this.closeConnection();
    }
    return res;
  } catch (error) {
    this.closeConnection();
    throw error;
  }
}

 private async exchange(request: PlistDictionary, timeout?: number): Promise<PlistDictionary> {
    const isNewConnection = !this.connection || this.isConnectionDestroyed();
    const conn = await this.connectToMobileImageMounterService();
    const res = await conn.sendPlistRequest(request, timeout);
    if (isNewConnection && res?.Request === 'StartService') {
      return await conn.receive();
    }
    return res;
  }

harsha509 (commented):
What do you think ?

src/services/ios/mobile-image-mounter/index.ts
navin772: makes sense, updated

harsha509 (approved):

PR #320, opened 2026-09-04
feat(companion-proxy): add CompanionProxyService for paired Apple Watch access

harsha509:
Introduces `CompanionProxyService` for `com.apple.companion_proxy.shim.remote`, the iPhone service that exposes its paired Apple Watch.

### Public API

| Method | Returns |
| --- | --- |
| `list()` | `string[]` — UDIDs of paired watches |
| `listen(timeout?)` | `AsyncGenerator<CompanionDeviceEvent>` — `GizmoPaired` / `GizmoUnpaired` / `GizmoAttach` / `GizmoDetach` |
| `getValue(companionUdid, key)` | `PlistValue` — e.g. `DeviceName`, `ProductVersion`, `BatteryCurrentCapacity` |
| `startForwardingServicePort(gizmoPort, options?)` | `number` — phone port proxying to the watch port |
| `stopForwardingServicePort(gizmoPort)` | — |
| `connectToForwardedPort(companionPort)` | `net.Socket` — raw TCP over the tunnel |
| `close()` | — |

Daemon errors are raised as `CompanionProxyError` with the raw `Error` string in `code`.

Wiring into `Services` and the package exports, plus the integration spec, follow in a separate PR.

### Testing

- `npm run build`, `npm run lint`, `npm run format:check`: clean
- `npm run test:unit`: 881 pass, 36 new
- Not yet verified against a real device with a paired watch

### DIFF
--- a/src/services/ios/companion-proxy/index.ts
+++ b/src/services/ios/companion-proxy/index.ts
@@ -104,36 +104,33 @@ export class CompanionProxyService extends BaseService implements CompanionProxy
     if (response.Error === 'NoPairedWatches') {
       return [];
     }
-    this.throwOnError(response, 'GetDeviceRegistry');
-    const udids = response.PairedDevicesArray;
+    const udids = this.throwOnError(response, 'GetDeviceRegistry').PairedDevicesArray;
     return Array.isArray(udids) ? udids.map(String) : [];
   }
 
   /**
    * Stream watch pair/unpair/attach/detach events on a dedicated connection that is closed
-   * when the consumer stops iterating, on error, or on `close()`.
+   * when the consumer stops iterating, on error, on `close()`, or when the daemon hangs up.
    * The daemon cancels the stream on any further input, so nothing else is ever sent on it.
    * @param timeout Milliseconds to wait for each event
    * @throws {CompanionProxyError} On a daemon `Error` reply or a frame that is not a companion event
-   * @throws {Error} When no event arrives within `timeout` or the connection is closed
+   * @throws {Error} When no event arrives within `timeout` or the connection is closed by either side
    */
   async *listen(timeout: number = DEFAULT_EVENT_TIMEOUT_MS): AsyncGenerator<CompanionDeviceEvent> {
     const conn = await this.connectToCompanionProxyService();
     this._listenConns.add(conn);
+    conn.getSocket().once('close', () => this.releaseListenConnection(conn));
     try {
       conn.sendPlist({Command: 'StartListeningForDevices'});
       while (true) {
-        const frame = await conn.receive(timeout);
-        this.throwOnError(frame, 'StartListeningForDevices');
+        const frame = this.throwOnError(await conn.receive(timeout), 'StartListeningForDevices');
         if (!isCompanionDeviceEvent(frame)) {
           throw new CompanionProxyError(`Unexpected companion event frame: ${JSON.stringify(frame)}`);
         }
         yield frame;
       }
     } finally {
-      if (this._listenConns.delete(conn)) {
-        this.closeListenConnection(conn);
-      }
+      this.releaseListenConnection(conn);
     }
   }
 
@@ -144,8 +141,10 @@ export class CompanionProxyService extends BaseService implements CompanionProxy
    * @throws {CompanionProxyError} On a daemon `Error` reply or a reply without `RetrievedValueDictionary`
    */
   async getValue(companionUdid: string, key: string): Promise<PlistValue> {
-    const response = await this.sendCommand(this.buildGetValueRequest(companionUdid, key));
-    this.throwOnError(response, 'GetValueFromRegistry');
+    const response = this.throwOnError(
+      await this.sendCommand(this.buildGetValueRequest(companionUdid, key)),
+      'GetValueFromRegistry',
+    );
     const values = response.RetrievedValueDictionary;
     if (!isPlistDictionary(values)) {
       throw new CompanionProxyError(`Unexpected GetValueFromRegistry reply: ${JSON.stringify(response)}`);
@@ -162,8 +161,10 @@ export class CompanionProxyService extends BaseService implements CompanionProxy
    * @throws {CompanionProxyError} On a daemon `Error` reply or a reply without `CompanionProxyServicePort`
    */
   async startForwardingServicePort(watchPort: number, options?: StartForwardingOptions): Promise<number> {
-    const response = await this.sendCommand(this.buildStartForwardingRequest(watchPort, options));
-    this.throwOnError(response, 'StartForwardingServicePort');
+    const response = this.throwOnError(
+      await this.sendCommand(this.buildStartForwardingRequest(watchPort, options)),
+      'StartForwardingServicePort',
+    );
     const port = response.CompanionProxyServicePort;
     if (typeof port !== 'number') {
       throw new CompanionProxyError(`Unexpected StartForwardingServicePort reply: ${JSON.stringify(response)}`);
@@ -179,8 +180,10 @@ export class CompanionProxyService extends BaseService implements CompanionProxy
    * @throws {CompanionProxyError} On a daemon `Error` reply or a reply that is not a success
    */
   async stopForwardingServicePort(watchPort: number): Promise<void> {
-    const response = await this.sendCommand(this.buildStopForwardingRequest(watchPort));
-    this.throwOnError(response, 'StopForwardingServicePort');
+    const response = this.throwOnError(
+      await this.sendCommand(this.buildStopForwardingRequest(watchPort)),
+      'StopForwardingServicePort',
+    );
     if (response.Command !== 'ComandSuccess' && response.Command !== 'CommandSuccess') {
       throw new CompanionProxyError(`Unexpected StopForwardingServicePort reply: ${JSON.stringify(response)}`);
     }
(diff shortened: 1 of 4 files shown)


src/services/ios/companion-proxy/index.ts
```diff
@@ -0,0 +1,267 @@
...
+
+function isPlistDictionary(value: PlistValue): value is PlistDictionary {
+  return util.isPlainObject(value);
+}
+
+function isCompanionDeviceEvent(frame: PlistDictionary): frame is CompanionDeviceEvent {
+  return typeof frame?.Command === 'string' && typeof frame.GizmoUDIDKey === 'string';
+}
+
+function assertGizmoPort(port: number): void {
```
mykola-mokhnach: it can still pass the same port value through on success

src/services/ios/companion-proxy/index.ts
```diff
@@ -0,0 +1,267 @@
...
+  }
+
+  /**
+   * Stream watch pair/unpair/attach/detach events on a dedicated persistent connection.
+   * The daemon cancels the stream on any further input, so nothing else is ever sent on it.
+   * @param timeout Milliseconds to wait for each event
+   */
+  async *listen(timeout: number = DEFAULT_EVENT_TIMEOUT_MS): AsyncGenerator<CompanionDeviceEvent> {
+    if (!this._listenConn) {
+      this._listenConn = await this.connectToCompanionProxyService();
```
mykola-mokhnach: this may potentially create a race

src/services/ios/companion-proxy/index.ts
```diff
@@ -0,0 +1,267 @@
...
+      this._listenConn.sendPlist({Command: 'StartListeningForDevices'});
+    }
+    const conn = this._listenConn;
+    while (true) {
+      const frame = await conn.receive(timeout);
+      this.throwOnError(frame, 'StartListeningForDevices');
```
mykola-mokhnach: is throwing an error the only way to exit this loop?

src/services/ios/companion-proxy/index.ts
```diff
@@ -0,0 +1,267 @@
...
+    if (typeof port !== 'number') {
+      throw new CompanionProxyError(`Unexpected StartForwardingServicePort reply: ${JSON.stringify(response)}`);
+    }
+    return port;
+  }
+
+  /**
+   * Stop forwarding a watch port started with `startForwardingServicePort()`.
+   * Accepts Apple's misspelled `ComandSuccess` reply as well as `CommandSuccess`.
+   * @param gizmoPort TCP port on the watch (not the phone port returned earlier)
```
mykola-mokhnach: not sure if we need to expose this internal `gizmo` prefix. It does not make any sense for external users

src/services/ios/companion-proxy/index.ts
```diff
@@ -0,0 +1,267 @@
...
+   */
+  async stopForwardingServicePort(gizmoPort: number): Promise<void> {
+    const response = await this.sendCommand(this.buildStopForwardingRequest(gizmoPort));
+    this.throwOnError(response, 'StopForwardingServicePort');
+    if (response.Command !== 'ComandSuccess' && response.Command !== 'CommandSuccess') {
+      throw new CompanionProxyError(`Unexpected StopForwardingServicePort reply: ${JSON.stringify(response)}`);
```
mykola-mokhnach, line 200: explicit method errors should be documented

src/services/ios/companion-proxy/index.ts
```diff
@@ -0,0 +1,267 @@
...
+  private buildStopForwardingRequest(gizmoPort: number): StopForwardingServicePortRequest {
+    assertGizmoPort(gizmoPort);
+    return {
+      Command: 'StopForwardingServicePort',
+      GizmoRemotePortNumber: gizmoPort,
+    };
+  }
+}
+
+export {CompanionProxyService};
```
mykola-mokhnach: cannot we just declare export together with class declaration?

src/services/ios/companion-proxy/index.ts
harsha509: Done, `assertWatchPort()` now returns the validated port and the builders use it inline.

src/services/ios/companion-proxy/index.ts
harsha509: Agree, Fixed it

src/services/ios/companion-proxy/index.ts
harsha509: wrapped the loop in `try/finally` so the connection is closed on every exit path.

src/services/ios/companion-proxy/index.ts
harsha509: agree, renamed `gizmoPort` to `watchPort` in the public API and error text

src/services/ios/companion-proxy/index.ts
harsha509: Added `@throws` tags for `CompanionProxyError` and `TypeError` on every public method in both class and interface

src/services/ios/companion-proxy/index.ts
harsha509: Done, switched to export class and dropped the trailing export statement.

src/services/ios/companion-proxy/index.ts
```diff
@@ -0,0 +1,286 @@
...
+    try {
+      conn.close();
+    } catch (error) {
+      log.error('Error closing companion proxy connection:', error);
+    }
+  }
+
+  private throwOnError(response: PlistDictionary, command: string): void {
+    if (!response?.Error) {
+      return;
```
mykola-mokhnach: this method may also return the original response in case of success

src/services/ios/companion-proxy/index.ts
```diff
@@ -0,0 +1,286 @@
...
+   * @param timeout Milliseconds to wait for each event
+   * @throws {CompanionProxyError} On a daemon `Error` reply or a frame that is not a companion event
+   * @throws {Error} When no event arrives within `timeout` or the connection is closed
+   */
+  async *listen(timeout: number = DEFAULT_EVENT_TIMEOUT_MS): AsyncGenerator<CompanionDeviceEvent> {
+    const conn = await this.connectToCompanionProxyService();
+    this._listenConns.add(conn);
+    try {
+      conn.sendPlist({Command: 'StartListeningForDevices'});
+      while (true) {
```
mykola-mokhnach, line 133: I still don't quite understand how this loop exits after we stop the service

src/services/ios/companion-proxy/index.ts
```diff
@@ -0,0 +1,286 @@
...
+  /**
+   * Stream watch pair/unpair/attach/detach events on a dedicated connection that is closed
+   * when the consumer stops iterating, on error, or on `close()`.
+   * The daemon cancels the stream on any further input, so nothing else is ever sent on it.
+   * @param timeout Milliseconds to wait for each event
+   * @throws {CompanionProxyError} On a daemon `Error` reply or a frame that is not a companion event
+   * @throws {Error} When no event arrives within `timeout` or the connection is closed
+   */
+  async *listen(timeout: number = DEFAULT_EVENT_TIMEOUT_MS): AsyncGenerator<CompanionDeviceEvent> {
+    const conn = await this.connectToCompanionProxyService();
```
mykola-mokhnach, line 125: what happens if `listen` is invoked twice in a row?

src/services/ios/companion-proxy/index.ts
harsha509: Done. `throwOnError` now returns the response on success, and all callers use the returned value directly.

src/services/ios/companion-proxy/index.ts
harsha509: Good catch, 

the previous fix only handled cleanup, not the exit itself. The pending `receive()` was only rejected on timeout or on our own `close()`, so if the daemon closed the stream the loop kept waiting until the event timeout.

Now `listen()` subscribes to the `socket close` event and releases the connection when it fires. That rejects the pending `receive()`, so the generator exits right away. 

Verified on a device.

src/services/ios/companion-proxy/index.ts
harsha509: Each call opens its own connection and its own `StartListeningForDevices` stream, so the two calls do not share any state. `close()` closes all of them. 

Verified on device that the daemon serves two listeners at the same time and both receive every event.

mykola-mokhnach: is such behavior expected? I assume close should be only affecting the instance it was created for, otherwise it would create unexpected side effects

harsha509: Hi @mykola-mokhnach ,

the listener set is per service instance, so `close()` on one `CompanionProxyService` never touches another instances streams. 

Within an instance, each `listen()` already closes its own connection when its consumer stops iterating, errors, or the daemon hangs up. close() is only the `dispose everything this instance holds` call, same as the other services.

If you would rather not have that cross-listener reach
1. Shalli Drop close() entirely; the service is otherwise stateless.
2. or shall i Keep it and add an `AbortSignal` option to `listen()` for per stream cancellation.

Happy to do either.

mykola-mokhnach: I would prefer # 2

harsha509: Updated !

mykola-mokhnach (approved):

PR #323, opened 2026-09-06
fix(remote-xpc): treat RST_STREAM and GOAWAY as fatal instead of ignoring them

harsha509:
The frame parser ignored RST_STREAM and GOAWAY frames, so when the device reset the connection the RSD connect attempt kept waiting until the post-handshake timeout. The transport now fails right away with the stream and error code. Adds tests for both frames and a PING control.

### DIFF
--- a/src/lib/remote-xpc/remote-xpc-framed-transport.ts
+++ b/src/lib/remote-xpc/remote-xpc-framed-transport.ts
@@ -6,7 +6,7 @@ import type {XPCDictionary} from '../types.js';
 import {Http2Constants} from './constants.js';
 import {DataFrame} from './handshake-frames.js';
 import Handshake from './handshake.js';
-import {Http2FrameParser, buildWindowUpdateFrames} from './http2-frame-parser.js';
+import {Http2FrameParser, type PeerTeardownFrame, buildWindowUpdateFrames} from './http2-frame-parser.js';
 import {decodeMessage, probeXpcFraming, XPC_WRAPPER_HEADER_SIZE} from './xpc-protocol.js';
 
 const log = getLogger('RemoteXpcFramedTransport');
@@ -272,6 +272,10 @@ export class RemoteXpcFramedTransport extends EventEmitter {
         this.flushPendingSends();
         continue;
       }
+      if (frame.type === 'rstStream' || frame.type === 'goAway') {
+        this.handlePeerTeardown(frame);
+        return;
+      }
       if (frame.type !== 'data') {
         continue;
       }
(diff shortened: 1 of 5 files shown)


src/lib/remote-xpc/remote-xpc-framed-transport.ts
```diff
@@ -356,10 +360,26 @@ export class RemoteXpcFramedTransport extends EventEmitter {
...
+
+  /**
+   * RST_STREAM and GOAWAY are fatal: the peer is closing the channel or the
+   * connection, and nothing usable follows either frame.
+   */
+  private handlePeerTeardown(frame: PeerTeardownFrame): void {
+    this.failConnection(
+      frame.type === 'rstStream'
+        ? `Peer sent RST_STREAM on stream ${frame.streamId} with error code ${frame.errorCode}`
+        : `Peer sent GOAWAY (last stream ${frame.lastStreamId}) with error code ${frame.errorCode}`,
```
mykola-mokhnach: do we always assume it will be GOAWAY as the opposite case?

src/lib/remote-xpc/remote-xpc-framed-transport.ts
mykola-mokhnach: switch would probably work better here

src/lib/remote-xpc/remote-xpc-framed-transport.ts
harsha509: good catch, The type only allows `RST_STREAM or GOAWAY`, but the ternary hid that, so I replaced it with an explicit switch

src/lib/remote-xpc/remote-xpc-framed-transport.ts
harsha509: Done, the loop now uses a switch

src/lib/remote-xpc/remote-xpc-framed-transport.ts
```diff
@@ -262,31 +267,42 @@ export class RemoteXpcFramedTransport extends EventEmitter {
...
-      }
-      if (frame.type !== 'data') {
-        continue;
+      switch (frame.type) {
+        case 'settings':
+          this.applyPeerSettings(frame.settings);
+          this.flushPendingSends();
+          break;
+        case 'windowUpdate':
+          this.adjustSendWindow(frame.streamId, frame.increment);
```
mykola-mokhnach: this block could be merged to the previous with a single `if`

src/lib/remote-xpc/remote-xpc-framed-transport.ts
```diff
@@ -356,10 +372,29 @@ export class RemoteXpcFramedTransport extends EventEmitter {
...
   private handleDesync(reason: string): void {
+    this.failConnection(reason);
+  }
+
+  /**
+   * RST_STREAM and GOAWAY are fatal: the peer is closing the channel or the
+   * connection, and nothing usable follows either frame.
+   */
+  private handlePeerTeardown(frame: PeerTeardownFrame): void {
+    switch (frame.type) {
```
mykola-mokhnach, line 384: should we add `default` handler?

mykola-mokhnach (approved):

src/lib/remote-xpc/remote-xpc-framed-transport.ts
harsha509: Merged the two cases with one flush call.

src/lib/remote-xpc/remote-xpc-framed-transport.ts
harsha509: Added a default that narrows the frame to never, so a new teardown type without a case fails to compile.

PR #327, opened 2026-09-15
fix(scripts): fall back to a free registry port when starting the Apple TV tunnel

harsha509:
## Problem

`start-appletv-tunnel.mjs` and `tunnel-creation.mjs` both default the tunnel registry to port 42314. Running the Apple TV script while the iOS tunnel script is already up fails immediately:

    Error: listen EADDRINUSE: address already in use 127.0.0.1:42314

The only workaround was to pass `--tunnel-registry-port` by hand.

## Fix

Add `scripts/lib/registry-port.mjs` with `resolveAvailableRegistryPort`, which probes the preferred port on 127.0.0.1 and walks up to the next free one (bounded to 20 ports), logging a warning when it substitutes. The Apple TV script uses it for the default port only. An explicit `--tunnel-registry-port` is honoured as given and still fails if that port is busy.

Note: the integration tests locate the registry through the persisted `tunnelRegistryPort` item, which the last started registry writes. With two registries up, the tests follow whichever started last.


scripts/lib/registry-port.mjs
```diff
@@ -0,0 +1,47 @@
...
+    probe.listen(port, TUNNEL_REGISTRY_HOST, () => probe.close(() => resolve(true)));
+  });
+}
+
+/**
+ * Ports from `preferredPort` upward, bounded by the attempt limit and the TCP range.
+ * @param {number} preferredPort
+ * @returns {number[]}
+ */
+function candidatePorts(preferredPort) {
```
mykola-mokhnach: This might take too long. Lets instead just use the port that the system advices: https://github.com/appium/appium/blob/dc16d6970f9806d46f92c41510bdeec4bee07702/packages/driver-test-support/lib/helpers.ts#L16

harsha509: Sure, updated.

mykola-mokhnach (approved):

PR #328, opened 2026-09-15
feat(amfi): add AmfiService for Developer Mode control

### LINKED ISSUE #325: feat: add `AmfiService` for Developer Mode control (`com.apple.amfi.lockdown.shim.remote`)
Add an `AmfiService` that talks to `com.apple.amfi.lockdown.shim.remote` over the RemoteXPC tunnel. The service controls Developer Mode on iOS 16+.

Developer Mode gates everything this library exists for: dev-signed app launch, debugger attachment, testmanagerd/XCTest, WebDriverAgent, and Developer Disk Image use. A fresh iOS 16+ device ships with it off and the Settings toggle hidden. Today users need Xcode to fix "Developer Mode is disabled" before Appium can start or change manually on device. This service lets the same Node process reveal or enable it. The read side already exists (`MobileImageMounterService.queryDeveloperModeStatus()`); this will help add the write side.

`pymobiledevice3` implementation - https://github.com/doronz88/pymobiledevice3/blob/master/pymobiledevice3/services/amfi.py but not implemented for RSD service.

## Protocol

Standard `.shim.remote` flow: `RSDCheckin`, drain the one unsolicited `{Request: 'StartService'}` greeting, then send a single plist `{action: N}`. Success reply is `{success: true}`; failures come back as `{Error: '...'}`.

| action | Effect |
|---|---|
| `0` | Reveal: makes the Developer Mode toggle visible under Settings > Privacy & Security. Does not change state. |
| `1` | Enable: turns Developer Mode on; the device reboots. Rejected with `{Error: 'Device has a passcode set'}` when a passcode exists. |
| `2` | Accept: answers the post-reboot "Turn on Developer Mode?" prompt so the flow can run unattended. |

## Proposed API

```ts
class AmfiService extends BaseService {
  static readonly RSD_SERVICE_NAME = 'com.apple.amfi.lockdown.shim.remote';
  revealDeveloperModeOption(): Promise<void>;   // action 0
  enableDeveloperMode(): Promise<void>;         // action 1, throws on Error
  acceptDeveloperModePrompt(): Promise<void>;   // action 2
}
```

## Testing
Test edge cases and combinations like device locked/unlocked, passcode set/unset, developer mode already on/off.

aayushisabharwal:
## Summary

Adds `AmfiService` for `com.apple.amfi.lockdown.shim.remote`, the write side of Developer Mode on iOS 16+. The read side already exists in `MobileImageMounterService.queryDeveloperModeStatus()`.

Closes #325

## API

```ts
class AmfiService extends BaseService {
  static readonly RSD_SERVICE_NAME = 'com.apple.amfi.lockdown.shim.remote';
  isDeveloperModeEnabled(): Promise<boolean>;   // strict status read, throws on failure
  revealDeveloperModeOption(): Promise<void>;   // action 0
  enableDeveloperMode(): Promise<void>;         // action 1, idempotent, throws DeviceHasPasscodeSetError
  acceptDeveloperModePrompt(): Promise<void>;   // action 2, idempotent
}
```

Exposed through `startAmfiService(udid)` in `services.ts`. `AmfiError`, `DeviceHasPasscodeSetError` and `DeveloperModeAction` are exported from the package root.

## Protocol details verified on hardware (iPhone 12, iOS 26)

| Action | Device state | Reply | Behaviour |
|---|---|---|---|
| 0 reveal | any | `{success: true}` | toggle shown in Settings; idempotent |
| 1 enable | off, passcode set | `{Error: 'Device has a passcode set'}` | `DeviceHasPasscodeSetError` |
| 1 enable | off, no passcode | `{success: 1}` (integer) | device reboots ~15 s later; after unlock the "Turn on Developer Mode?" alert is pending and status still reads off |
| 1 enable | **already on** | `{success: 1}` | AMFI still reboots and comes back with Developer Mode **off** until confirmed. The service reads the status first and sends nothing in this case |
| 2 accept | alert pending | success | Developer Mode on, no further reboot |
| 2 accept | nothing pending | `{Error: 'An unknown error has occured'}` (sic), or the passcode error on a locked device | surfaced as `AmfiError` with that `code`; no-op when Developer Mode is already on |

Notes for reviewers:

- After `RSDCheckin` the shim sends one unsolicited `{Request: 'StartService'}` frame, drained before the request. Same pattern as `CompanionProxyService`.
- `success` is a boolean for action 0 but the integer `1` for action 1. Both are accepted.
- Each call opens a fresh checked-in connection and closes it, matching pymobiledevice3.
- Reboot-and-reconnect orchestration is intentionally out of scope. The enable reboot tears down the tunnel; `scripts/tunnel-creation.mjs` reconnects only when started with `--reconnect-retries 0`.

## Testing

- `npm run test:unit`: 28 new unit tests under `test/unit/amfi/` using a fake connection. They cover the greeting drain, the action integers, idempotence with and without a passcode, the passcode and other error paths, and the integer success reply.
- `npm run test:amfi`: integration spec. Status and reveal run by default. `AMFI_RUN_ENABLE=1` and `AMFI_RUN_ACCEPT=1` opt in to the state-changing steps: run enable, let the device reboot, unlock it, restart the tunnel, then run accept. A passcode makes the enable step fail with `DeviceHasPasscodeSetError` by design.


src/services/ios/amfi/index.ts
```diff
@@ -0,0 +1,180 @@
...
+import type {AmfiService as AmfiServiceInterface, PlistDictionary, PlistValue} from '../../../lib/types.js';
+import type {ServiceConnection} from '../../../service-connection.js';
+import {BaseService} from '../base-service.js';
+import {AmfiError, DeviceHasPasscodeSetError} from './errors.js';
+
+const log = getLogger('AmfiService');
+
+const DEFAULT_TIMEOUT_MS = 10000;
+/** Shim that answers `QueryDeveloperModeStatus`; read before enabling so enable stays idempotent */
+const MOBILE_IMAGE_MOUNTER_SERVICE_NAME = 'com.apple.mobile.mobile_image_mounter.shim.remote';
```
navin772: We could import instead of re-declaring - `MobileImageMounterService.RSD_SERVICE_NAME` to keep service names consistent.

src/services.ts
```diff
@@ -102,6 +103,14 @@ export async function startPowerAssertionService(udid: string): Promise<PowerAss
   return new PowerAssertionService(udid);
 }
 
+/**
+ * Start the AMFI service (Developer Mode control) for the given device UDID.
+ */
+export async function startAmfiService(udid: string): Promise<AmfiService> {
+  await requireCatalogService(udid, AmfiService.RSD_SERVICE_NAME);
```
navin772, line 110: We should check for image mounter service too as we are using it for Amfi:
```ts
await requireCatalogService(udid, MobileImageMounterService.RSD_SERVICE_NAME);
```

src/services/ios/amfi/index.ts
```diff
@@ -0,0 +1,180 @@
...
+    this.timeout = timeout;
+  }
+
+  /**
+   * Read the current Developer Mode state through the mobile image mounter shim.
+   * Unlike `MobileImageMounterService.queryDeveloperModeStatus()`, a failed read throws instead of
+   * reporting "enabled", so `enableDeveloperMode()` never skips on a transport error.
+   */
+  async isDeveloperModeEnabled(): Promise<boolean> {
+    const response = await this.exchange(MOBILE_IMAGE_MOUNTER_SERVICE_NAME, {Command: 'QueryDeveloperModeStatus'});
```
navin772: Can we directly use `QueryDeveloperModeStatus` from image mounter instead of `this.exchange`?

src/services/ios/amfi/index.ts
aayushisabharwal: Done in 1108b57: imports `MobileImageMounterService.RSD_SERVICE_NAME` instead of re-declaring the shim name.

src/services.ts
aayushisabharwal: Done in 1108b57: `startAmfiService` now also requires the image mounter shim in the catalog.

src/services/ios/amfi/index.ts
aayushisabharwal: Thought about that, but queryDeveloperModeStatus() swallows every error and returns true, so a failed read would make enable/accept think dev mode is already on and silently skip. Kept a strict read here on purpose. Happy to add a throwing variant on the mounter and use that instead if you'd prefer.

navin772 (approved):
LGTM!

harsha509 (approved):
LGTM!

harsha509:
Good Work @aayushisabharwal ! 🎉

PR #331, opened 2026-09-18
fix(xctest): pass `ActivateSuspended: true` to background the WDA

navin772:
RemoteXPC fix for https://github.com/appium/appium-xcuitest-driver/issues/2978

Xcuitest fix - https://github.com/appium/appium-xcuitest-driver/pull/2980


harsha509 (approved):
LGTM!

PR #334, opened 2026-09-21
fix(plist): parse 1/2/4-byte integers as unsigned

harsha509:
## Problem

Binary plist 1/2/4-byte integers are unsigned per CoreFoundation, but the parser read them with `readInt8/16/32BE`. Any value above the signed maximum decoded negative.

Surfaced by a device-supplied ephemeral port:

```
Forwarded watch 62078 -> phone -10952     # 54584 read as signed 16-bit
```

The bug needs a value **≥ 32768 that the sender encoded in 2 bytes**, which is why it stayed hidden: most plist integers are small, and large ones (e.g. `TotalDataCapacity`) land in the correctly-signed 8-byte path. Kernel-assigned ephemeral ports (49152–65535) hit it on most draws.

Our own writer promoted such values to 4 bytes, so round-tripping through it never reproduced the bug. Apple's encoder packs them in 2.

## Fix

- **Parser** — read the 1/2/4-byte forms unsigned; leave the 8-byte form signed.
- **Writer** — the short forms previously carried negatives, which would now decode wrong, so every negative routes to the 8-byte branch. Negatives still round-trip; shorter widths stay unambiguous.

Both sides are required: the parser change alone would break negatives.


src/lib/plist/binary-plist-creator.ts
```diff
@@ -218,20 +218,16 @@ class BinaryPlistCreator {
...
-      buffer.writeUInt8(BPLIST_TYPE.INT | 0, 0);
-      buffer.writeInt8(value, 1);
-    } else if (value >= -32768 && value <= 32767) {
+    } else if (value >= 0 && value <= 65535) {
       buffer = Buffer.alloc(3);
       buffer.writeUInt8(BPLIST_TYPE.INT | 1, 0);
-      buffer.writeInt16BE(value, 1);
-    } else if (value >= -2147483648 && value <= 2147483647) {
+      buffer.writeUInt16BE(value, 1);
+    } else if (value >= 0 && value <= 4294967295) {
```
mykola-mokhnach: should we compare the lower bound to 65535 here? would it make sense to use hex representation for better readability?

mykola-mokhnach (approved):

src/lib/plist/binary-plist-creator.ts
harsha509: agreed. 
Updated

PR #335, opened 2026-09-21
feat(companion-proxy): register service and add integration tests

harsha509:
## What

`CompanionProxyService` was implemented and unit-tested but had **no factory in `services.ts`**, so nothing outside the package could reach it:

```
Services.startCompanionProxyService   undefined
Services.startNotificationProxyService  function   # for comparison
```

Adds `startCompanionProxyService`, mirroring the existing factories (`requireCatalogService` + construct).

## Integration spec

Six tests against a real iPhone with a paired Apple Watch:

- enumerate paired watches
- read registry values (`DeviceName`, `ProductVersion`, `UniqueDeviceID`)
- identity consistency — the registry's own `UniqueDeviceID` must equal the UDID `list()` returned, plus `DeviceClass` and `ProductType`
- live battery and storage telemetry, with `AmountDataAvailable <= TotalDataCapacity`
- reject a well-formed but unpaired UDID with `NoMatchingWatch`
- forward watch port `62078` through the phone and TCP-connect to the returned port

Port `62078` is deliberate: watch lockdownd is the one port exempt from the daemon's source-IP pin. The forward is released in a `finally`, including on assertion failure.


src/lib/plist/binary-plist-creator.ts
```diff
@@ -218,20 +218,16 @@ class BinaryPlistCreator {
       buffer = Buffer.alloc(2);
       buffer.writeUInt8(BPLIST_TYPE.INT | 0, 0);
       buffer.writeUInt8(value, 1);
-    } else if (value >= -128 && value <= 127) {
-      buffer = Buffer.alloc(2);
-      buffer.writeUInt8(BPLIST_TYPE.INT | 0, 0);
-      buffer.writeInt8(value, 1);
-    } else if (value >= -32768 && value <= 32767) {
+    } else if (value >= 0 && value <= 65535) {
```
mykola-mokhnach: lets merge the other pr and then rebase to avoid repetitions

mykola-mokhnach (approved):

PR #336, opened 2026-09-22
feat(usbmux): add listen() for device attach/detach notifications

mikebruin:
## Summary

`Usbmux` only implements one-shot request/response calls (`ListDevices`, `ReadBUID`, ...). usbmuxd's wire protocol also has a `Listen` request that streams `Attached`/`Detached` notifications on the same connection for as long as it stays open. This PR adds support for it to the library only. The tunnel-creation script changes that use it have moved to a follow-up PR, as suggested in review.

- `src/lib/usbmux/index.ts`: add `Usbmux#listen({signal?})`, which returns an async iterable of `UsbmuxAttachEvent | UsbmuxDetachEvent`. usbmuxd reports every already-connected device as an attach right after acknowledging the request. usbmuxd has no way to unsubscribe, so listening takes over the connection:
  - only one `listen()` is allowed per connection
  - calling it on a closed connection throws
  - the connection is closed once iteration stops (`break`, `return`, abort, or `close()`)
  - `close()` is now idempotent
- `src/lib/usbmux/device-event-stream.ts`: the queue/wake/stop state now lives in a dedicated `UsbmuxDeviceEventStream` class instead of closures inside `listen()`.
- `Attached`/`Detached` frames are routed by `MessageType`, not by request tag, because usbmuxd sends them unsolicited with tag 0 (see libimobiledevice's `notify_device_add`/`notify_device_remove`). Routing by tag only worked when `Listen` happened to be the first request on the connection.
- Frames go to a pending request's callback first. Only frames that no request is waiting for are treated as Listen notifications. Before the plist parser fix in #295 (v5.14.5), nested `DeviceList` entries were flattened, so a `ListDevices` reply carried a top-level `MessageType: 'Attached'`. Checking pending requests first means a reply is never taken for an event. This matters for the fix backported onto older releases.
- If the usbmuxd connection drops (e.g. usbmuxd restarts), `listen()` iteration now rejects with `usbmuxd connection closed`, after delivering any events already received, so callers can resubscribe. Before, it waited forever on a dead socket. Concurrent `next()` calls are all resolved in turn.
- `src/index.ts`: export the `UsbmuxAttachEvent` / `UsbmuxDetachEvent` / `UsbmuxDeviceEvent` types.

## Test plan

- [x] `npm run build`, `npm run lint`, `npm run format:check`: all clean
- [x] `test/unit/usbmux/usbmux-listen.spec.ts` (16 tests) against a mock usbmuxd built from the real `UsbmuxEncoder` wire format. The mock sends notifications with tag 0 like real usbmuxd. Covered: attach, detach, ordered multi-event delivery, abort via signal, an already-aborted signal, `close()` ending iteration, a dropped connection, a non-dictionary plist root, Listen failure, Listen not being the first request on the connection, and a reply carrying `MessageType: 'Attached'` still reaching its pending request. Also covered: a second `listen()` throwing, the connection closing after `break`, queued events delivered before a drop error, concurrent `next()` calls, and `listen()` on a closed connection throwing.
- [x] `npm run test:unit`: 957/957 pass
- [ ] Linux / Windows: not tested yet (details in the review reply)
- [x] Ran against real usbmuxd on a macOS 27 host with 14 USB-connected iPhones, sending `ListDevices`/`ReadBUID` first so Listen wasn't tag 0:
  - `listDevices()` returned 14 devices, and `listen()` then delivered 14 well-formed USB attach events matching that list.
  - The previous revision of this PR received 0 events in the same setup.
  - Also verified backported onto 5.14.4, where the reply-routing fix above is required.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


scripts/tunnel-creation.mjs
```diff
@@ -33,6 +33,12 @@ const log = logger.getLogger('TunnelCreation');
 /** @type {import('appium-ios-remotexpc').TunnelRegistryServer | null} */
 let registryServer = null;
 
+/** @type {import('appium-ios-remotexpc').Usbmux | null} */
```
mykola-mokhnach, line 36: xcuitest has its own tunnel creation script, which means this logic has to be ported there as soon as this change is merged and released

scripts/tunnel-creation.mjs
```diff
@@ -562,6 +657,12 @@ async function main() {
       const firstUdid = successful[0].device.Properties.SerialNumber;
       log.info(`   curl "http://localhost:${registryPort}/remotexpc/tunnels/${firstUdid}?waitMs=15000"`);
     }
+
+    if (!specificUdid) {
+      // Runs for the remaining lifetime of the process (the registry server's open
+      // socket is what actually keeps the daemon alive; this just rides alongside it).
+      await watchForHotplugDevices({devicesByUdid, reconnectTunnelByUdid});
```
mykola-mokhnach, line 664: I assume this logic should be properly orchestrated together with existing device reconnection logic (perhaps, we would also need different command line arguments). Also, not all devices use usbmux as their transport.

mykola-mokhnach: maybe it even makes sense to split both features (e.g. listen support in usbmux and reconnection logic in the tunnel creation script) into different dependable PRs

src/lib/usbmux/index.ts
```diff
@@ -193,7 +219,109 @@ export class Usbmux extends BaseSocketService {
...
+    return {
+      [Symbol.asyncIterator](): AsyncIterator<UsbmuxDeviceEvent> {
+        const tag = self._tag++;
+        const queue: UsbmuxDeviceEvent[] = [];
+        let wake: (() => void) | null = null;
+        let stopped = false;
+        let failure: Error | null = null;
+        let gotInitialResult = false;
+
+        const notifyWaiter = () => {
```
mykola-mokhnach, line 251: would it make more sense to isolate these closures into a separate class?

scripts/tunnel-creation.mjs
mikebruin: Understood. The script changes have moved to #337, which puts the watcher in one self-contained `watchDevices()` function so it's straightforward to port. I'll open the matching `appium-xcuitest-driver` change once this is released.

scripts/tunnel-creation.mjs
mikebruin: Agreed. In #337, hot-plug watching is now opt-in behind `--watch-devices`, so default behaviour is unchanged. It shares the reconnect state:
- attaches are ignored while a tunnel exists or a reconnect is in progress for that UDID
- a detach prunes `devicesByUdid`, which `runReconnectAttempts` now checks before each attempt, so it stops retrying a device that is gone
- losing one of a device's USB/Network connections isn't treated as a detach

The help text and README say it only covers usbmux-visible devices.

mikebruin: Done. This PR now contains only `Usbmux#listen()` plus its tests. The script changes are in #337, which depends on this one.

src/lib/usbmux/index.ts
mikebruin: Done: it's now `UsbmuxDeviceEventStream` in `src/lib/usbmux/device-event-stream.ts`. While doing this I found that usbmuxd sends `Attached`/`Detached` with tag 0 rather than the Listen request's tag, so routing by tag only worked when Listen was the first request on the connection. Notifications are now routed by `MessageType`, and there's a test for it.

src/lib/usbmux/device-event-stream.ts
```diff
@@ -0,0 +1,123 @@
...
+  private _failure: Error | null = null;
+  private readonly _signal?: AbortSignal;
+  private readonly _onStop: (stream: UsbmuxDeviceEventStream) => void;
+  private readonly _onAbort = () => this.stop();
+
+  /**
+   * @param onStop - Invoked exactly once when the stream stops, so the owner can unregister it
+   * @param signal - When aborted, stops the stream
+   */
+  constructor(onStop: (stream: UsbmuxDeviceEventStream) => void, signal?: AbortSignal) {
```
mykola-mokhnach: add dedicated type alias for the onStop type

src/lib/usbmux/device-event-stream.ts
```diff
@@ -0,0 +1,123 @@
...
+    this._onStop = onStop;
```
mykola-mokhnach: can we make these to property shortcuts? Starting underscore is not really necessary for private properties

src/lib/usbmux/device-event-stream.ts
```diff
@@ -0,0 +1,123 @@
...
+    if (this._stopped) {
+      return;
+    }
+    this._stopped = true;
+    this._signal?.removeEventListener('abort', this._onAbort);
+    this._onStop(this);
+    this._notifyWaiter();
+  }
+
+  async next(): Promise<IteratorResult<UsbmuxDeviceEvent>> {
```
mykola-mokhnach: consider adding jsdoc to all public methods

src/lib/usbmux/index.ts
```diff
@@ -90,7 +94,15 @@ export class Usbmux extends BaseSocketService {
 
     this._tag = 0;
     this._responseCallbacks = {};
+    this._eventStreams = new Set();
```
mykola-mokhnach: default property values could also be set upon definition

src/lib/usbmux/index.ts
```diff
@@ -283,6 +341,21 @@ export class Usbmux extends BaseSocketService {
     const handler = this._responseCallbacks[data.header.tag];
     if (handler) {
       handler(data);
+      return;
+    }
+
+    // Listen notifications arrive unsolicited with tag 0, not the Listen request's tag. Only
+    // frames no pending request is waiting for are treated as such, so a ListDevices reply
+    // (whose entries also carry MessageType 'Attached') is never mistaken for one.
+    const {MessageType} = data.payload;
```
mykola-mokhnach: is it safe to destruct like that? is payload always an object?

src/lib/usbmux/device-event-stream.ts
mikebruin: Done: added `UsbmuxDeviceEventStreamStopHandler` in e33112e.

src/lib/usbmux/device-event-stream.ts
mikebruin: Done: `onStop` and `signal` are now `private readonly` constructor parameter properties. I also dropped the underscore prefixes from the other private members of `UsbmuxDeviceEventStream`.

src/lib/usbmux/device-event-stream.ts
mikebruin: Done: every public member of `UsbmuxDeviceEventStream` now has JSDoc (`stopped`, `push`, `fail`, `stop`, `next`, `return`, `[Symbol.asyncIterator]`).

src/lib/usbmux/index.ts
mikebruin: Done: `_eventStreams` is now initialized where it's declared. I kept the underscore so it matches the other `Usbmux` fields.

src/lib/usbmux/index.ts
mikebruin: Good catch, it wasn't safe. `UsbmuxDecoder` only casts the `parsePlist()` result to `PlistDictionary`. The root can be an array, a scalar or null, and destructuring null would throw inside the decoder's `data` handler. There's now an `asPlistDictionary()` guard, used both here and in the Listen result callback. Frames without a dictionary root are ignored, and there's a test for it.

src/lib/usbmux/index.ts
```diff
@@ -90,7 +94,15 @@ export class Usbmux extends BaseSocketService {
 
     this._tag = 0;
     this._responseCallbacks = {};
+    this._eventStreams = new Set();
     this._decoder.on('data', this._handleData.bind(this));
+    // Unless close() already stopped them, surface a dropped connection (e.g. usbmuxd restarting)
+    // to listen() consumers instead of leaving them waiting forever
+    this._socketClient.on('close', () => {
+      for (const stream of [...this._eventStreams]) {
```
mykola-mokhnach: this pattern repeats multiple times (e.g. for stream of this._eventStreams)
maybe it makes sense to extract it to a tiny private wrapper, which accepts a tiny lambda callback to reduce the amount of boilerplate code?

mikebruin: Done in 2539460: added a private `_forEachEventStream(fn)` helper. The connection-close handler, `close()` and notification dispatch in `_handleData` now all go through it. It iterates over a snapshot of the set, because stopping a stream removes it from the set.

mykola-mokhnach (approved):

harsha509:
Thanks for this. A few things I noticed:

1. Calling `listen()` twice on the same `Usbmux` sends a second `Listen`. usbmuxd sends the device list again, so the first listener gets duplicate attach events. Can we throw on a second `listen()` for the same connection?
2. `break` or abort stops the stream but leaves the socket open. usbmuxd has no unlisten command, so the connection stays in listen mode. Can we close the connection when the stream stops?
3. When the connection drops, `next()` throws before returning events that are already queued. A detach that arrived just before the drop gets lost. Can we return the queued events first and throw after that?
4. Two `next()` calls at the same time leave the first one waiting forever, because `wake` only holds one resolver. A list of waiters would fix it.
5. Calling `listen()` on a `Usbmux` whose socket is already closed waits forever. It should fail right away.
6. The comment in `_handleData` says a ListDevices reply can be mistaken for an event. That was a plist parser bug fixed in #295 (v5.14.5), so the comment can be updated.

Also, has this been tested on Linux and Windows?

mikebruin:
@harsha509 thanks, all six were real. They're fixed in 4eb3441, with a test for each of 1–5:

1. **Second `listen()` on the same connection** now throws (`listen() is already active on this usbmuxd connection; use a separate connection`).
2. **Stopping the stream closes the connection.** `break`, `return()` and abort all close it, since usbmuxd has no unlisten. `close()` is now memoized, so it's safe when both the stream and the caller close the connection.
3. **Dropped connection:** `next()` now returns events that were already queued first, and only then rejects with `usbmuxd connection closed`.
4. **Concurrent `next()` calls:** the single `wake` resolver is now a list of waiters, and every pending call is resolved in turn.
5. **`listen()` on a closed connection** now throws right away (`Cannot listen on a closed usbmuxd connection`).
6. **The `_handleData` comment** now says the `ListDevices` collision came from the parser bug fixed in #295 (v5.14.5). I kept the "pending requests first" ordering as a safeguard. It costs nothing, and it's needed wherever this fix gets backported onto releases older than v5.14.5. We run one on 5.14.4.

`npm run test:unit` passes 957/957. #337 is rebased onto this. Its watcher already gave each `listen()` its own connection and closed it on the way out, so no script changes were needed.

**Linux / Windows:** not tested yet. Everything so far has been on macOS: the unit tests, plus a macOS 27 host with 14 USB-connected iPhones running against the real usbmuxd.
- **Linux:** the connection code there is the same Unix-socket path (`/var/run/usbmuxd`) as on macOS. The "notifications arrive with tag 0" behaviour comes from libimobiledevice's usbmuxd source (`notify_device_add` / `notify_device_remove` send with tag 0), and that is the usbmuxd Linux uses. So I expect it to work, but haven't run it.
- **Windows:** usbmux goes through Apple Mobile Device Service over TCP 27015. I haven't verified how it tags Listen notifications. The routing doesn't depend on a specific tag value (anything no pending request is waiting for counts as a notification), so it should hold up, but that's unverified.

I'll run the same real-device check on Linux when I can get a box with a device attached. If anyone has Windows access, a quick run would be very welcome.

src/lib/usbmux/index.ts
```diff
@@ -219,20 +220,32 @@ export class Usbmux extends BaseSocketService {
...
+    }
+    if (this._eventStreams.size > 0) {
+      throw new Error('listen() is already active on this usbmuxd connection; use a separate connection');
+    }
+
     const tag = this._tag++;
     const stream = new UsbmuxDeviceEventStream((s) => {
       delete this._responseCallbacks[tag];
       this._eventStreams.delete(s);
+      void this.close().catch(() => {});
```
harsha509: we dont want to silence errors. can we log here

src/lib/usbmux/index.ts
```diff
@@ -219,20 +220,32 @@ export class Usbmux extends BaseSocketService {
...
     }, opts.signal);
     if (stream.stopped) {
+      void this.close().catch(() => {});
```
harsha509: same as above

harsha509 (approved):

src/lib/usbmux/index.ts
mikebruin: Done in 915f9a0: both call sites now go through a private `_closeAfterListen()`, which logs failures with `log.warn` instead of swallowing them. I did the same for the two `close().catch(() => {})` calls in #337's script.

src/lib/usbmux/index.ts
mikebruin: Same fix: this call site also uses `_closeAfterListen()` now (915f9a0).

harsha509:
Thank you @mikebruin !

